PR #14, opened 2020-06-25
Make the HTTP Client usage easier

satish860:
This PR contains the changes to make the Meilisearch easy to use.

PR #16, opened 2020-07-01
Change and implement documents methods

curquiza:
- rename `Meilisearch` into `MeiliSearch` in comments
- rename `AddOrUpdateDocuments` into `UpdateDocuments`
- implement `AddDocuments`

I want to merge #14 before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ NB: you can also download MeiliSearch from **Homebrew** or **APT**.
 ```c#
 MeilisearchClient client = new MeilisearchClient("http://localhost:7700", "masterKey");
 var index = await client.CreateIndex("movies");
-var updateStatus = await index.AddorUpdateDocuments<Movie>(new Movie[] {new Movie {Id = "1", Name = "Batman"}, new Movie{Id="2",Name = "Interstellar"}});
+var updateStatus = await index.AddDocuments<Movie>(new Movie[] {new Movie {Id = "1", Name = "Batman"}, new Movie{Id="2",Name = "Interstellar"}});
 Movie movie = await index.GetDocument<Movie>("1");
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -99,14 +99,17 @@ var indexes = await client.GetAllIndexes();
 ```
 ### Documents
 
-#### Add Documents <!-- omit in toc -->
+#### Add or Update Documents <!-- omit in toc -->
 
 ```c#
- var updateStatus = await index.AddorUpdateDocuments(new[]{new  Movie {Id = "1", Name = "Batman"}});
+ var updateStatus = await index.AddDocuments(new[]{new  Movie {Id = "1", Name = "Batman"}});
+ var updateStatus = await index.UpdateDocuments(new[]{new  Movie {Id = "1", Name = "Batman"}});
```
satish860, line 112: I thought based on the ID , Meilisearch supports adding or Update.

satish860 (approved):

README.md
curquiza: You're right it supports both but in different ways!
In the MeiliSearch API there are:
- the “Add or replace documents” route: it adds a document if it does not exist and overwrite the previous document otherwise
- the “Add or update documents” route: it adds a document if it does not exist and partially update the previous document otherwise (it does not overwrite the previous fields if not mentioned in the body)

Why do we choose to call our method (in the SDKs) `addDocuments` for `addOrReplaceDocuments` and `updateDocuments` for `AddOrUpdateDocuments`? For simplicity. It’s clearer for the user without reading the MeiliSearch docs. You want to add: you use `addDocuments`. You want to update: you use `updateDocuments`. An API has to be accurate in its routes naming. An SDK has to be easy to use since it's a wrapper to help user play and code with MeiliSearch 🚀
That’s why we decided to keep these shorter names 🙂

PR #27, opened 2020-07-07
Use the default client in tests since it's the default usage

curquiza:


README.md
```diff
@@ -202,6 +204,19 @@ var movies = await this.index.Search<Movie>("ironman");
...
 ```
 
+## 🧰 Use a Custom HTTP Client
+
+You can replace the default client used in this package by the one you want.
+
+For example:
+
+```c#
+var _httpClient = ClientFactory.Instance.CreateClient<MeilisearchClient>();
```
satish860, line 215: This example might need an update. The way we have created Meilisearch , It can be used as a Typed HttpClient. More details here: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-3.1#typed-clients

curquiza: Hum I'm not sure to really understand. Can you write down the example you would like to see in the README instead of this one? 🙂

satish860: Sure will do

PR #31, opened 2020-07-08
Update Document Methods to accept Ints or strings for document ids

### LINKED ISSUE #25: Documents methods should accept an int or a string
With:

```c#
public class Movie
{
    public int Id { get; set; }
}
```

```c#
MeilisearchClient client = new MeilisearchClient("http://localhost:7700", "masterKey");
var index = client.CreateIndex("movies");
var documents = new Movie[] {
    new Movie {Id = 1, Name = "Batman"}
};
var updateStatus = await index.AddDocuments<Movie>(documents);
Movie movie = await index.GetDocument<Movie>(1);
```

I got the error "impossible to convert int into a string".
The `GetDocument` method should be able to accept an int or a string since MeiliSearch can accept a string or an int as a document identifier.

TODO:
- [ ] Fix in the code
- [ ] Add tests

barbados-clemens:
Updates the following methods to have an overload for int documentId(s)
- `GetDocument<T>`
- `DeleteOneDocument`
- `DeleteDocuments`

These methods call the string variant in order to be DRY. 
Tests added in fashion of the string variant method call.

closes #25

### DIFF
--- a/src/Meilisearch/Index.cs
+++ b/src/Meilisearch/Index.cs
@@ -107,6 +107,11 @@ namespace Meilisearch
            return await this._client.GetFromJsonAsync<T>($"/indexes/{Uid}/documents/{documentId}");
         }
 
+        public async Task<T> GetDocument<T>(int documentId)
+        {
+            return await this._client.GetFromJsonAsync<T>($"/indexes/{Uid}/documents/{documentId}");
+        }
+
         /// <summary>
         /// Get documents with the allowed Query Parameters.
         /// </summary>
--- a/src/Meilisearch/Index.cs
+++ b/src/Meilisearch/Index.cs
@@ -174,12 +174,8 @@ namespace Meilisearch
         /// <returns>Update status with ID to look for progress of update.</returns>
         public async Task<UpdateStatus> DeleteDocuments(IEnumerable<int> documentIds)
         {
-            // We could do
-            // var docIds = documentIds.Select(id => id.ToString())
-            // await this.DeleteDocuments(docIds)
-            // which would keep keep the code more DRY, but at the cost of casting all to a strings
-            var httpresponse = await this._client.PostAsJsonAsync($"/indexes/{Uid}/documents/delete-batch", documentIds);
-            return await httpresponse.Content.ReadFromJsonAsync<UpdateStatus>();
+            var docIds = documentIds.Select(id => id.ToString());
+            return await this.DeleteDocuments(docIds);
         }
 
         /// <summary>
--- a/src/Meilisearch/Index.cs
+++ b/src/Meilisearch/Index.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Json;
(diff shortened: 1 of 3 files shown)


src/Meilisearch/Index.cs
```diff
@@ -134,6 +146,16 @@ public async Task<UpdateStatus> DeleteOneDocument(string documentId)
             return await httpresponse.Content.ReadFromJsonAsync<UpdateStatus>();
         }
 
+        /// <summary>
+        /// Delete one document by its ID
+        /// </summary>
+        /// <param name="documentId">document ID</param>
+        /// <returns>Update Status with ID to look for document.</returns>
+        public async Task<UpdateStatus> DeleteOneDocument(int documentId)
+        {
```
satish860, line 155: @barbados-clemens Awesome work on the PR.  Just want to understand your perspective on a different approach.  
What if we create a class class `Identity` which has to implement an `implicit `C# operator which can convert both Int and string. So we don't need two methods.
More details here  : https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/user-defined-conversion-operators

satish860 (changes requested):
I would like to understand If use of `implicit` operator can help here.

src/Meilisearch/Index.cs
barbados-clemens: I'm not familiar with with implicit/explicit operators, but it sounds promising 
> Implicit conversions don't require special syntax to be invoked and can occur in a variety of situations, for example, in assignments and methods invocations

What would that look like in practice?

if I understand the documentation correctly, Their example just wraps the type `byte` and We'd need to wrap int and string? Does this mean the sdk methods would take `Identity` instead of an int or string?
i.e. 
` public async Task<T> GetDocument<T>(Identity documentId)`

What would our Identity class look like? I think that's the part I'm unsure about since I'm not understanding exactly how the provided example would translate in this case.

satish860: That's the wrong suggestion from my side. Just found out that we cannot add an implicit operator on the Framework defined class. 
https://stackoverflow.com/questions/17785487/implicit-operator-for-int

barbados-clemens:
After debugging this Test it seems the error is coming out of [MeiliSearchClient.cs Line 67](https://github.com/meilisearch/meilisearch-dotnet/blob/master/src/Meilisearch/MeilisearchClient.cs#L67). 

I didn't touch this file and it was working fine before I resolved the merge confilcts. All the Document tests, which would indicate the merge conflict was resolved correctly. 

The error specifically is `System.Threading.Tasks.TaskCanceledException: The operation was canceled.`

So something is timing out. Looking at the log I don't ever see the POST request for creating a new index arrive. 

image


Looking through the commit history I don't see anything that actually touches the MeiliSearchClient.cs file since I started my branch. 

I'm going to keep debugging to see if I can figure out what happened.

barbados-clemens:
Okay when running these tests isolated, they pass. so that's not a good sign. 

Edit: Now it all passes locally when run via the built in Rider test runner, but fail when running via the commandline, `dotnet test`

but it's failing on (IndexTests.cs line 14](https://github.com/meilisearch/meilisearch-dotnet/blob/master/tests/Meilisearch.Tests/IndexTests.cs#L18) now, essentially any test that calls [CreateIndex](https://github.com/meilisearch/meilisearch-dotnet/blob/master/src/Meilisearch/MeilisearchClient.cs#L64)

Any ideas @satish860 going forward?

Edit 2: reverting commit _Auto Update dosettings.user (#25)_ ([commit link](https://github.com/barbados-clemens/meilisearch-dotnet/commit/942b1100bed3c6d3f742f6de91394db8b0c984e1)). Solved the issue running `dotnet test` locally, all tests pass, but still are failing in the pipeline.

satish860 (approved):

curquiza:
@barbados-clemens I don't know why the tests failed, I re-ran them and now it works. If it fails again, we should open an issue about this random failure! Sorry for the time you spent on this problem! 😕

curquiza:
I merge since @satish860 approved! Thanks @barbados-clemens for your work and @satish860 for the review! 🎉

PR #48, opened 2020-08-06
Remove useless dependency

curquiza:
Since this package is not used, I remove it.

Do I miss something @satish860?
Is this package used somewhere?


satish860:
Coverlet is used here to get code coverage number . If we remove it then we might not get code coverage

curquiza:
@satish860 how can I get the code coverage? 🙂 I've never seen it

satish860:
Since we don't use Cake build, This is not needed.

satish860 (commented):
We can add it back when we start collecting the coverage. As of now, I am good with removal.

satish860 (dismissed):

curquiza:
Perfect! We indeed put i back when needed 🙂 Merging!

PR #52, opened 2020-08-20
Add linter and format check

### LINKED ISSUE #9: Linter
We should add a linter to this project, at least for the code style.

For example, I can see here the same kind of method but with not the same style (especially in the name of the function):

https://github.com/meilisearch/meilisearch-dotnet/blob/086c738d2254435687e7f45db2996fd7826d11d4/tests/Meilisearch.Tests/DocumentTests.cs#L23-L43

The goal is to provide a better environment for contribution 🙂

I'm not a .NET dev, so I'm not aware of the lint tools in .NET. What do you suggest?

NB: I'm adding an editorconfig file in this PR #8 but it's just about general style indentation.

curquiza:
- add [`dotnet-format`](https://github.com/dotnet/format) check in CI
- remove DotSettings file, see [why](https://stackoverflow.com/questions/22234617/do-i-need-to-maintain-resharpers-dotsettings-user-file-in-source-control).
- add StyleCop
- modify tests to avoid random failure

Fixes #9


satish860:
This is awesome . Never seen this tool

PR #59, opened 2020-09-10
Add GetOrCreateIndex method

curquiza:
Gets the index instance or creates a new one if it does not exists.

### DIFF
--- a/src/Meilisearch/MeilisearchClient.cs
+++ b/src/Meilisearch/MeilisearchClient.cs
@@ -104,7 +104,9 @@ namespace Meilisearch
 
         /// <summary>
         /// Gets the index instance or creates the index if it does not exist.
-        /// /!\ Really basics. An error handler should be created to check the exception.
+        ///
+        /// /!\ Really basics. An error handler should be created to check the errorCode.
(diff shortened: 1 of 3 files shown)


satish860 (commented):
Throwing an exception and handling will be costly. Is there a http status code on create which we can depend on in create and send a error object based on that.

curquiza:
MeiliSearch returns a `4XX` if the index already exists and indeed returns an error code. But we need to implement a better error handler first to handle that, see the comment you added in `CreateIndex`:
https://github.com/meilisearch/meilisearch-dotnet/blob/f00af958fc30c5c1bcc8e57faed646c1e3fe19ce/src/Meilisearch/MeilisearchClient.cs#L71-L72
As I noted, this is a first and beta version of `GetOrCreateIndex` (only for tests purpose for the moment). We'll improve it when an error handler will be implemented 🙂 So, for now, we need to catch the exception

curquiza:
I merge this PR since I need to keep working on this package 🙂 This function is mainly for tests purpose and will be improved once the error handler with be present.

@satish860 if you see any big issue, feel free to open a PR or an issue about that! 😉

PR #68, opened 2020-10-13
Improve tests

### LINKED ISSUE #46: Remove indexes after running tests
We should erase the indexes present in the MeiliSearch instance before starting the tests: this is dangerous because we run the test in a context we don't know.
This issue is not a problem on the CI, but is not really convenient when running tests locally.

curquiza:
- add synonyms tests
- refacto the structure of tests
- fixes #46


eskombro (approved):
👌 Amazing work @curquiza !

THANKS!!! 🌮 for you!

PR #86, opened 2020-12-07
Set SearchQuery.Filters type as string

### LINKED ISSUE #85: Set SearchQuery.Filters type to be a string
The MeiliSearch documentation specify that the SearchQuery.Filters property is of type string and not array of string : https://docs.meilisearch.com/guides/advanced_guides/search_parameters.html#filters

The associated type of this property should be changed and test updated to cover this. Without this change, it's impossible to use this field (at least with the version 0.17.0 that i have on my laptop).

Note : I'll take time to do a PR in order to fix that.

XREvo:
This PR is relative to issue #85 (Set SearchQuery.Filters type to be a string)
It containes :
- Changed SearchQuery.Filters type to be a string
- Added tests to cover its usage

### DIFF
--- a/src/Meilisearch/SearchQuery.cs
+++ b/src/Meilisearch/SearchQuery.cs
@@ -23,9 +23,9 @@ namespace Meilisearch
         public int? Limit { get; set; }
 
         /// <summary>
-        /// Gets or sets filters to apply to the query.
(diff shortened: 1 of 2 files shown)


curquiza (approved):
Awesome! Thanks for adding this test, this is exactly what we needed! ❤️ Merging!

bors merge

PR #111, opened 2021-04-14
Revert "Added NbHits and ExhaustiveNbHits to SearchResult<T> object."

curquiza:
Revert changes pushed on `staging`


curquiza:
useless since staging was already cleaned by bors, sorry

PR #117, opened 2021-05-03
Adding Error Handler Dotnet

### LINKED ISSUE #30: A basic error handler
We should add a basic error handler, with these 2 custom errors:
- `MeiliSearchApiError`
- `MeiliSearchCommunicationError`

If I'm not wrong, @barbados-clemens you created one in your previous meilisearch-dotnet repo (not available anymore) 🙂
Was this complicated to add?

FYI: Even if it's not documented yet, MeiliSearch API has a great error handler, and returns this kind of JSON when an error is found:

Capture d’écran 2020-07-02 à 10 18 21

It could be great to let the user access `message`, `errorCode`, `errorType` and `errorLink`.

alallema:
**Description**
Implement the error handler by using `HttpMessageHandler` and adapt to the error handler of other SDKs with 3 main errors type:
- `MeiliSearchApiError`
- `MeiliSearchCommunicationError`
- `MeiliSearchTimeoutError`

**Issue related**
#30

### DIFF
--- a/src/Meilisearch/Errors/MeilisearchApiError.cs
+++ b/src/Meilisearch/Errors/MeilisearchApiError.cs
@@ -10,7 +10,7 @@ namespace Meilisearch
     {
         /// <summary>
         /// Initializes a new instance of the <see cref="MeilisearchApiError"/> class.
-        /// Handler Exception received from Meilisearch API.
+        /// Handler Exception received from MeiliSearch API.
         /// </summary>
         /// <param name="apiError">Specific error message from Meilisearch Api.</param>
         public MeilisearchApiError(MeilisearchApiErrorContent apiError)
--- a/src/Meilisearch/Errors/MeilisearchApiError.cs
+++ b/src/Meilisearch/Errors/MeilisearchApiError.cs
@@ -1,0 +1,38 @@
+namespace Meilisearch
+{
+    using System;
+    using System.Net;
+
+    /// <summary>
+    /// Error sent by MeiliSearch API.
+    /// </summary>
+    public class MeilisearchApiError : Exception
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MeilisearchApiError"/> class.
+        /// Handler Exception received from Meilisearch API.
+        /// </summary>
+        /// <param name="apiError">Specific error message from Meilisearch Api.</param>
+        public MeilisearchApiError(MeilisearchApiErrorContent apiError)
+            : base(string.Format("MeilisearchApiError, Message: {0}, ErrorCode: {1}, ErrorType: {2}, ErrorLink: {3}", apiError.Message, apiError.ErrorCode, apiError.ErrorType, apiError.ErrorLink))
+        {
+            this.ErrorCode = apiError.ErrorCode;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MeilisearchApiError"/> class.
+        /// Handler Exception when Meilisearch API doesn't send a response message.
+        /// </summary>
+        /// <param name="statusCode">Status code from http response message.</param>
+        /// <param name="reasonPhrase">Reason Phrase from http response message.</param>
+        public MeilisearchApiError(HttpStatusCode statusCode, string reasonPhrase)
+            : base(string.Format("MeilisearchApiError, Message: {0}, ErrorCode: {1}", reasonPhrase, (int)statusCode))
+        {
+        }
+
+        /// <summary>
+        /// Gets or sets the errorCode return by MeilisearchApi..
+        /// </summary>
+        public string ErrorCode { get; set; }
+    }
+}
--- a/src/Meilisearch/Errors/MeilisearchApiError.cs
+++ b/src/Meilisearch/Errors/MeilisearchApiError.cs
@@ -31,7 +31,7 @@ namespace Meilisearch
         }
 
         /// <summary>
-        /// Gets or sets the errorCode return by MeilisearchApi..
+        /// Gets or sets the errorCode return by MeilisearchApi.
         /// </summary>
         public string ErrorCode { get; set; }
     }
--- a/src/Meilisearch/Errors/MeilisearchApiErrorContent.cs
+++ b/src/Meilisearch/Errors/MeilisearchApiErrorContent.cs
@@ -1,0 +1,34 @@
+namespace Meilisearch
+{
+    using System.Text.Json.Serialization;
+
+    /// <summary>
+    /// Error sent by MeiliSearch API.
+    /// </summary>
+    public class MeilisearchApiErrorContent
+    {
+        /// <summary>
+        /// Gets or sets the message.
+        /// </summary>
+        [JsonPropertyName("message")]
+        public string Message { get; set; }
+
+        /// <summary>
+        /// Gets or sets the errorCode.
+        /// </summary>
+        [JsonPropertyName("errorCode")]
+        public string ErrorCode { get; set; }
+
+        /// <summary>
+        /// Gets or sets the errorType.
+        /// </summary>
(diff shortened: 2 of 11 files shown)


src/Meilisearch/Errors/MeilisearchTimeoutError.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MeilisearchTimeoutError"/> class.
+        /// Handler Exception for MeilisearchTimeoutError with message and inner exception.
+        /// </summary>
+        /// <param name="message">Custom error message.</param>
+        /// <param name="innerException">Inner exception.</param>
+        public MeilisearchTimeoutError(string message, Exception innerException)
+            : base(string.Format("MeilisearchTimeoutError, ", message), innerException)
+        {
+        }
```
curquiza: Are both methods useful?

src/Meilisearch/MeilisearchMessageHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// </summary>
+        /// <param name="innerHandler">InnerHandler.</param>
+        public MeilisearchMessageHandler(HttpMessageHandler innerHandler)
+            : base(innerHandler)
+        {
+        }
+
+        /// <summary>
+        /// Try to override SendAsync.
+        /// BEWARE : Throws error if the index already exist. Use GetIndex before using Create.
```
curquiza: Did you add this line @alallema or did you copy/paste from another file? I'm not sure to get the contexte

curquiza (dismissed):

src/Meilisearch/MeilisearchClient.cs
```diff
@@ -115,18 +107,16 @@ public async Task<Index> GetOrCreateIndex(string uid, string primaryKey = defaul
...
             }
-            catch (Exception e)
+            catch (MeilisearchApiError e)
             {
-                if (e.Message == "Not able to create index. May be Index already exist")
-                {
-                    return await this.GetIndex(uid);
-                }
-                else
+                if (e.Content.ErrorCode != "index_not_found")
```
curquiza: Can we simplify the usage with `e.ErrorCode`?

src/Meilisearch/Errors/MeilisearchApiErrorContent.cs
```diff
@@ -0,0 +1,34 @@
+namespace Meilisearch
```
curquiza, line 1: I think it complexifies to separate the MeilISearchApiError into `MeilisearchApiErrorContent.cs` and `MeilISearchApiError.cs`
Can we remain it simple?

src/Meilisearch/Errors/MeilisearchApiError.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MeilisearchApiError"/> class.
+        /// Handler Exception received by Meilisearch API with inner exception.
+        /// </summary>
+        /// <param name="apiError">Specific error message from Meilisearch Api.</param>
+        public MeilisearchApiError(MeilisearchApiErrorContent apiError)
+            : base(string.Format("MeilisearchApiError, Message: {0}, ErrorCode: {1}, ErrorType: {2}, ErrorLink: {3}", apiError.Message, apiError.ErrorCode, apiError.ErrorType, apiError.ErrorLink))
+        {
+            this.Content = apiError;
+        }
```
curquiza: I didn't check but I'm worried the MeiliSearchApiError will not work if the body does not contain the `errorCode`, `errorType` ...
Indeed, it could happen in the case NGINX catches the error before MeiliSearch and returns a 502, for example -> the body will not be formatted as expected and it will raise an exception in your exception.

You should add a test like I did in the dart SDK with a wrong URL -> the MeiliSearch server will not return the "nice" mody with all the parameters: https://github.com/meilisearch/meilisearch-dart/blob/b9f0526736b7fc9baae8781a4fc5741212970698/test/exceptions_test.dart#L21-L29

curquiza (changes requested):

src/Meilisearch/MeilisearchMessageHandler.cs
alallema: I stole it indeed!

curquiza: because it's not true anymore 😁

src/Meilisearch/Errors/MeilisearchTimeoutError.cs
alallema: No, the second one isn't useful anymore.

src/Meilisearch/Errors/MeilisearchApiErrorContent.cs
alallema: I have some trouble merging both classes due to a collision between
`public virtual string Message { get; }`  in `Exception` parent class and
`[JsonPropertyName("message")]`
`        public string MeilisearchApiMessage { get; set; }` from Meilisearch Api
I will try to find a way

curquiza: Checked with @alallema, we did not find an easy way at the moment. We let it as it is.

src/Meilisearch/MeilisearchClient.cs
alallema: Done

src/Meilisearch/Errors/MeilisearchApiError.cs
alallema: You right I change the code to catch this possibility and add a test.

curquiza (commented):
Waiting for the update for the custom client, but everything that has been done so far looks perfect 😍

curquiza (dismissed):
Bravooo 👏 
LGTM!

src/Meilisearch/Errors/MeilisearchApiError.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// </summary>
+        /// <param name="statusCode">Status code from http response message.</param>
+        /// <param name="reasonPhrase">Reason Phrase from http response message.</param>
+        public MeilisearchApiError(HttpStatusCode statusCode, string reasonPhrase)
+            : base(string.Format("MeilisearchApiError, Message: {0}, ErrorCode: {1}", reasonPhrase, (int)statusCode))
+        {
+        }
+
+        /// <summary>
+        /// Gets or sets the errorCode return by MeilisearchApi..
```
curquiza:
```suggestion
        /// Gets or sets the errorCode return by MeilisearchApi.
```

src/Meilisearch/Errors/MeilisearchCommunicationError.cs
```diff
@@ -0,0 +1,21 @@
+namespace Meilisearch
+{
+    using System;
+
+    /// <summary>
+    /// Error sent when trying to connecting to Meilisearch.
```
curquiza:
```suggestion
    /// Error sent when trying to connecting to MeiliSearch.
```

src/Meilisearch/Errors/MeilisearchApiError.cs
```diff
@@ -0,0 +1,38 @@
...
+    using System.Net;
+
+    /// <summary>
+    /// Error sent by MeiliSearch API.
+    /// </summary>
+    public class MeilisearchApiError : Exception
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MeilisearchApiError"/> class.
+        /// Handler Exception received from Meilisearch API.
```
curquiza:
```suggestion
        /// Handler Exception received from MeiliSearch API.
```

curquiza (approved):
bors merge

PR #118, opened 2021-05-11
Changes due to the implicit index creation

alallema:
**Description**
Implicit index creation
- New function:
`Index(string uid)`
`FetchPrimaryKey()`
- Changing behavior function:
`AddDocument(documents, string primaryKey = default)`
`UpdateDocument(documents, string primaryKey = default)`

### DIFF
--- a/src/Meilisearch/Index.cs
+++ b/src/Meilisearch/Index.cs
@@ -40,6 +40,18 @@ namespace Meilisearch
         public string PrimaryKey { get; internal set; }
 
         /// <summary>
+        /// Fetch the primary key of the index.
+        /// </summary>
+        /// <returns>Primary key of the index fetch.</returns>
+        public async Task<string> FetchPrimaryKey()
+        {
+            var response = await this.client.GetAsync($"indexes/{this.Uid}");
+            var content = await response.Content.ReadFromJsonAsync<Index>();
+            this.PrimaryKey = content.PrimaryKey;
+            return this.PrimaryKey;
+        }
+
+        /// <summary>
         /// Changes the primary key of the index.
         /// </summary>
         /// <param name="primarykeytoChange">Primary key set.</param>
@@ -67,11 +79,20 @@ namespace Meilisearch
         /// Add documents.
         /// </summary>
         /// <param name="documents">Documents to add.</param>
+        /// <param name="primaryKey">Primary key for the documents.</param>
         /// <typeparam name="T">Type of the document. Even though documents are schemaless in MeiliSearch, making it typed helps in compile time.</typeparam>
         /// <returns>Returns the updateID of this async operation.</returns>
-        public async Task<UpdateStatus> AddDocuments<T>(IEnumerable<T> documents)
+        public async Task<UpdateStatus> AddDocuments<T>(IEnumerable<T> documents, string primaryKey = default)
         {
-            var responseMessage = await this.client.PostAsJsonAsync($"/indexes/{this.Uid}/documents", documents);
+            HttpResponseMessage responseMessage;
+            string uri = $"/indexes/{this.Uid}/documents";
+            if (primaryKey != default)
+            {
+                uri = QueryHelpers.AddQueryString(uri, new { primaryKey = primaryKey }.AsDictionary());
+            }
+
+            responseMessage = await this.client.PostAsJsonAsync(uri, documents);
+
             return await responseMessage.Content.ReadFromJsonAsync<UpdateStatus>();
         }
 
@@ -79,11 +100,20 @@ namespace Meilisearch
         /// Update documents.
         /// </summary>
         /// <param name="documents">Documents to update.</param>
+        /// <param name="primaryKey">Primary key for the documents.</param>
         /// <typeparam name="T">Type of document. Even though documents are schemaless in MeiliSearch, making it typed helps in compile time.</typeparam>
         /// <returns>Returns the updateID of this async operation.</returns>
-        public async Task<UpdateStatus> UpdateDocuments<T>(IEnumerable<T> documents)
+        public async Task<UpdateStatus> UpdateDocuments<T>(IEnumerable<T> documents, string primaryKey = default)
         {
-            var responseMessage = await this.client.PutAsJsonAsync($"/indexes/{this.Uid}/documents", documents);
+            HttpResponseMessage responseMessage;
+            string uri = $"/indexes/{this.Uid}/documents";
+            if (primaryKey != default)
+            {
+                uri = QueryHelpers.AddQueryString(uri, new { primaryKey = primaryKey }.AsDictionary());
+            }
+
+            responseMessage = await this.client.PutAsJsonAsync(uri, documents);
+
             return await responseMessage.Content.ReadFromJsonAsync<UpdateStatus>();
         }
 
--- a/src/Meilisearch/Index.cs
+++ b/src/Meilisearch/Index.cs
@@ -54,7 +54,7 @@ namespace Meilisearch
         /// <summary>
         /// Fetch the primary key of the index.
         /// </summary>
-        /// <returns>Primary key of the index fetch.</returns>
+        /// <returns>Primary key of the fetched index.</returns>
         public async Task<string> FetchPrimaryKey()
         {
             return (await this.FetchInfo()).PrimaryKey;
--- a/src/Meilisearch/Index.cs
+++ b/src/Meilisearch/Index.cs
@@ -40,15 +40,24 @@ namespace Meilisearch
         public string PrimaryKey { get; internal set; }
 
         /// <summary>
+        /// Fetch the info of the index.
+        /// </summary>
+        /// <returns>An instance of the index fetch.</returns>
+        public async Task<Index> FetchInfo()
+        {
+            var response = await this.client.GetAsync($"indexes/{this.Uid}");
+            var content = await response.Content.ReadFromJsonAsync<Index>();
+            this.PrimaryKey = content.PrimaryKey;
+            return this;
+        }
+
+        /// <summary>
         /// Fetch the primary key of the index.
         /// </summary>
         /// <returns>Primary key of the index fetch.</returns>
         public async Task<string> FetchPrimaryKey()
         {
-            var response = await this.client.GetAsync($"indexes/{this.Uid}");
-            var content = await response.Content.ReadFromJsonAsync<Index>();
-            this.PrimaryKey = content.PrimaryKey;
-            return this.PrimaryKey;
+            return (await this.FetchInfo()).PrimaryKey;
         }
 
         /// <summary>
--- a/src/Meilisearch/MeilisearchClient.cs
+++ b/src/Meilisearch/MeilisearchClient.cs
@@ -104,10 +104,7 @@ namespace Meilisearch
         /// <returns>Returns Index or Null if the index does not exist.</returns>
         public async Task<Index> GetIndex(string uid)
         {
-            var response = await this.client.GetAsync($"/indexes/{uid}");
-            var content = await response.Content.ReadFromJsonAsync<Index>();
-
-            return content.WithHttpClient(this.client);
+            return await this.Index(uid).FetchInfo();
         }
 
         /// <summary>
--- a/src/Meilisearch/MeilisearchClient.cs
+++ b/src/Meilisearch/MeilisearchClient.cs
@@ -61,7 +61,7 @@ namespace Meilisearch
         /// Calling this method doesn't create an index in the MeiliSearch instance, but grants access to all the other methods in the Index class.
         /// </summary>
         /// <param name="uid">Unique Id.</param>
-        /// <returns>Returns Index.</returns>
+        /// <returns>Returns an Index instance.</returns>
         public Index Index(string uid)
         {
             Index index = new Index(uid);
--- a/src/Meilisearch/MeilisearchClient.cs
+++ b/src/Meilisearch/MeilisearchClient.cs
@@ -45,7 +45,7 @@ namespace Meilisearch
         }
 
         /// <summary>
-        /// Gets the current MeiliSearch version. For more details on response
+        /// Gets the current MeiliSearch version. For more details on response.
         /// https://docs.meilisearch.com/reference/api/version.html#get-version-of-meilisearch.
         /// </summary>
         /// <returns>Returns the MeiliSearch version with commit and build version.</returns>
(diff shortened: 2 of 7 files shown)


tests/Meilisearch.Tests/IndexTests.cs
```diff
@@ -9,11 +11,187 @@ namespace Meilisearch.Tests
...
+        public async Task IndexCreationWithPrimaryKey()
+        {
+            var indexUid = "IndexCreationWithPrimaryKeyTest";
+            var index = await this.defaultClient.CreateIndex(indexUid, this.defaultPrimaryKey);
+            index.Uid.Should().Be(indexUid);
+            index.PrimaryKey.Should().Be(this.defaultPrimaryKey);
+        }
+
+        [Fact]
+        public async Task BasicImpliciteIndexCreation()
```
curquiza: Why is this test named "implicit index creation"?
The implicit index creation is a feature of MeiliSearch: MeiliSearch creates the index when the users push the documents if the index does not already exist.
I realized it's not clear because of the title of the main issue in Integration guides 😅

tests/Meilisearch.Tests/IndexTests.cs
```diff
@@ -9,11 +11,187 @@ namespace Meilisearch.Tests
...
+        public async Task FetchPrimaryKey()
+        {
+            var indexUid = "FetchPrimaryKeyTest";
+            var index = this.defaultClient.Index(indexUid);
+            index.Uid.Should().Be(indexUid);
+            index.PrimaryKey.Should().BeNull();
+
+            var document = await index.AddDocuments(new[] { new { movieId = "1", Name = "Batman" } }, this.defaultPrimaryKey);
+
+            Thread.Sleep(50);
```
curquiza: Can you use waitForPengindUpdate please?

tests/Meilisearch.Tests/IndexTests.cs
curquiza: Here you only test the `Index()` method

tests/Meilisearch.Tests/IndexTests.cs
```diff
@@ -9,11 +11,187 @@ namespace Meilisearch.Tests
...
+            var indexUid = "ImpliciteIndexCreationWithPrimaryKeyTest";
+            var index = this.defaultClient.Index(indexUid);
+            index.Uid.Should().Be(indexUid);
+            var document = await index.AddDocuments(new[] { new Movie { Id = "1", Name = "Batman" } }, this.defaultPrimaryKey);
+            index.PrimaryKey.Should().Be(this.defaultPrimaryKey);
+            document.UpdateId.Should().BeGreaterOrEqualTo(0);
+        }
+
+        [Fact]
+        public async Task ImpliciteIndexCreationOnExistingIndex()
```
curquiza: This test is not an implicit index creation. It looks like you test `Index()` on an already existing index. No need to add docs in this test.

tests/Meilisearch.Tests/IndexTests.cs
```diff
@@ -9,11 +11,187 @@ namespace Meilisearch.Tests
...
+            index = this.defaultClient.Index(indexUid);
+            index.Uid.Should().Be(indexUid);
+
+            var document = await index.AddDocuments(new[] { new Movie { Id = "1", Name = "Batman" } });
+
+            document.UpdateId.Should().BeGreaterOrEqualTo(0);
+        }
+
+        [Fact]
+        public async Task ImpliciteIndexCreationOnExistingIndexWithPrimaryKey()
```
curquiza: Same, this is not implicit index creation, the index is already created. Looks like this test is to check the validity of the primary key when adding document -> should be in the DocumentsTest.cs file 🙂

src/Meilisearch/Index.cs
```diff
@@ -67,23 +79,49 @@ public async Task<bool> Delete()
...
+            if (primaryKey == default)
+            {
+                responseMessage = await this.client.PostAsJsonAsync(uri, documents);
+            }
+            else
+            {
+                this.PrimaryKey = primaryKey;
+                uri = QueryHelpers.AddQueryString(uri, new { primaryKey = primaryKey }.AsDictionary());
+                responseMessage = await this.client.PostAsJsonAsync(uri, documents);
+            }
```
curquiza:
```suggestion
            if (primaryKey != default)
            {
                uri = QueryHelpers.AddQueryString(uri, new { primaryKey = primaryKey }.AsDictionary());
            }
            responseMessage = await this.client.PostAsJsonAsync(uri, documents);
```

curquiza: We should not add the primary key as the new primary key of the index (`this.PrimaryKey`) since the user can pass something that will not be taken into account by MeiliSearch, eg if the primary key is already defined.

src/Meilisearch/Index.cs
```diff
@@ -67,23 +79,49 @@ public async Task<bool> Delete()
...
-            var responseMessage = await this.client.PutAsJsonAsync($"/indexes/{this.Uid}/documents", documents);
+            HttpResponseMessage responseMessage;
+            string uri = $"/indexes/{this.Uid}/documents";
+            if (primaryKey == default)
+            {
+                responseMessage = await this.client.PutAsJsonAsync(uri, documents);
+            }
+            else
+            {
+                this.PrimaryKey = primaryKey;
```
curquiza: same here: don't update the primary key of the index instance

curquiza (changes requested):
Thanks A LOT for this PR! ❤️ 

Some remarks:
- "Implicit" does not spell with a "e" at the end in English 🙃
- Can you add tests of `AddDocuments` and `UpdateDocuments` to check we can pass a primary key when using the method? In the [DocumentsTest.cs](https://github.com/meilisearch/meilisearch-dotnet/blob/main/tests/Meilisearch.Tests/DocumentTests.cs) file?

tests/Meilisearch.Tests/IndexTests.cs
alallema: You totally right is because I don't want made confusion with IndexCreation but I will rename this test it's not clear at all

alallema: Done

tests/Meilisearch.Tests/IndexTests.cs
alallema: Actually, this test checks the validity of the primary key on the existing index so I remove `AddDocument()` and I let it in IndexTest.cs file

src/Meilisearch/Index.cs
alallema: Done

tests/Meilisearch.Tests/IndexTests.cs
alallema: Done its work! Actually, I don't set the good primaryKey it needs to be in lower case to be set properly.
My key was never take

tests/Meilisearch.Tests/IndexTests.cs
alallema: I use `AddDocument()` in this test to check the Index object works even if the Index was already created with `CreateIndex`. Maybe I can use another function like settings or stats?

tests/Meilisearch.Tests/IndexTests.cs
```diff
@@ -9,11 +10,179 @@ namespace Meilisearch.Tests
...
+        }
+
+        [Fact]
+        public async Task ImplicitIndexCreationWithPrimaryKey()
+        {
+            var indexUid = "ImplicitIndexCreationWithPrimaryKeyTest";
+            var index = this.defaultClient.Index(indexUid);
+            index.Uid.Should().Be(indexUid);
+            var document = await index.AddDocuments(new[] { new Movie { Id = "1", Name = "Batman" } }, this.defaultPrimaryKey);
+            index.PrimaryKey.Should().BeNull();
```
curquiza: Here you don't test an implicit index creation because you don't test the index is created.
You should:
- use a primary key that does not contain `id` in the name (otherwise MeiliSearch is able to infer)
- wait for the document addition to end
- check you can get the index and the primary key is the good one
- do this test of implicit index creation with a document addition without primary key

Anyway I'm not sure this test should be added in the IndexTests.cs file, it's as you wish.

tests/Meilisearch.Tests/IndexTests.cs
curquiza: Use `FetchPrimaryKey` instead :)

tests/Meilisearch.Tests/DocumentTests.cs
```diff
@@ -50,19 +69,37 @@ public async Task BasicDocumentsAdditionWithTimeoutError()
...
 
+        [Fact]
+        public async Task DocumentsAdditionWithPrimaryKey()
+        {
+            var indexUid = "DocumentsAdditionWithPrimaryKeyTest";
+            var index = this.client.Index(indexUid);
+            index.PrimaryKey.Should().BeNull();
+
+            // Add the documents
+            var document = await index.AddDocuments(new[] { new { movieId = "1", Name = "Batman" } }, "movieId");
```
curquiza: Use a primary key that does not contain `id`

tests/Meilisearch.Tests/DocumentTests.cs
```diff
@@ -89,6 +126,24 @@ public async Task BasicDocumentsUpdate()
...
 
+        [Fact]
+        public async Task DocumentsUpdateWithPrimaryKey()
+        {
+            var indexUid = "DocumentsUpdateWithPrimaryKeyTest";
+            var index = this.client.Index(indexUid);
+            index.PrimaryKey.Should().BeNull();
+
+            // Add the documents
+            var document = await index.UpdateDocuments(new[] { new { movieId = "1", Name = "Batman" } }, "movieId");
```
curquiza: Use a primary key that does not contain id

tests/Meilisearch.Tests/IndexTests.cs
```diff
@@ -9,11 +10,179 @@ namespace Meilisearch.Tests
...
+        {
+            var indexUid = "CheckValidityOfPrimaryKeyOnExistingIndexTest";
+            var createIndex = await this.defaultClient.CreateIndex(indexUid, this.defaultPrimaryKey);
+            var indexObject = this.defaultClient.Index(indexUid);
+            Assert.Equal(createIndex.Uid, indexObject.Uid);
+            Assert.Equal(this.defaultPrimaryKey, createIndex.PrimaryKey);
+            indexObject.PrimaryKey.Should().BeNull();
+            await indexObject.FetchPrimaryKey();
+            Assert.Equal(createIndex.PrimaryKey, indexObject.PrimaryKey);
+        }
```
curquiza:
```suggestion
```

src/Meilisearch/Index.cs
```diff
@@ -39,6 +39,18 @@ public Index(string uid, string primaryKey = default)
         /// </summary>
         public string PrimaryKey { get; internal set; }
 
+        /// <summary>
+        /// Fetch the primary key of the index.
+        /// </summary>
+        /// <returns>Primary key of the index fetch.</returns>
+        public async Task<string> FetchPrimaryKey()
+        {
+            var response = await this.client.GetAsync($"indexes/{this.Uid}");
```
curquiza, line 48: use `getIndex`

tests/Meilisearch.Tests/IndexTests.cs
alallema: I deleted it

tests/Meilisearch.Tests/DocumentTests.cs
alallema: Done

tests/Meilisearch.Tests/DocumentTests.cs
alallema: Done

tests/Meilisearch.Tests/IndexTests.cs
alallema: This is the function I delete

alallema:
@curquiza we talk to implement `Update()` method but this function already exist:
```
        public async Task<Index> ChangePrimaryKey(string primarykeytoChange)
        {
            var message = await this.client.PutAsJsonAsync($"indexes/{this.Uid}", new { primaryKey = primarykeytoChange });
            var responsecontent = await message.Content.ReadFromJsonAsync<Index>();
            this.PrimaryKey = responsecontent.PrimaryKey;
            return this;
        }
```
I can rename it for more coherence between the SDKs but I find this name more accurate maybe we can change it to `UpdatePrimaryKey()` ?

curquiza:
@alallema, thanks for the investigation!
You can rename it in another PR 🙂 I will open an issue

Request me for a review when you think this is ready :)

src/Meilisearch/Index.cs
alallema: I added the `FetchInfo()` method like in others SDKs and use it in the `GetIndex()` method too.

src/Meilisearch/Index.cs
```diff
@@ -39,6 +39,27 @@ public Index(string uid, string primaryKey = default)
...
+            var content = await response.Content.ReadFromJsonAsync<Index>();
+            this.PrimaryKey = content.PrimaryKey;
+            return this;
+        }
+
+        /// <summary>
+        /// Fetch the primary key of the index.
+        /// </summary>
+        /// <returns>Primary key of the index fetch.</returns>
```
curquiza:
```suggestion
        /// <returns>Primary key of the fetched index.</returns>
```

src/Meilisearch/MeilisearchClient.cs
```diff
@@ -56,6 +56,19 @@ public async Task<MeiliSearchVersion> GetVersion()
             return await response.Content.ReadFromJsonAsync<MeiliSearchVersion>();
         }
 
+        /// <summary>
+        /// Create a local reference to an index identified by UID, without doing an HTTP call.
+        /// Calling this method doesn't create an index in the MeiliSearch instance, but grants access to all the other methods in the Index class.
+        /// </summary>
+        /// <param name="uid">Unique Id.</param>
+        /// <returns>Returns Index.</returns>
```
curquiza:
```suggestion
        /// <returns>Returns an Index instance.</returns>
```

tests/Meilisearch.Tests/DocumentTests.cs
```diff
@@ -1,6 +1,7 @@
 namespace Meilisearch.Tests
 {
     using System.Linq;
+    using System.Threading;
```
curquiza:
```suggestion
```

curquiza (changes requested):
Some final requests before the final merge 😇

alallema:
> Some final requests before the final merge 😇

Done!

curquiza (approved):
Bravo @alallema 👏 

bors merge

PR #129, opened 2021-06-03
Fix partial update of settings

### LINKED ISSUE #67: The UpdateDocuments method should be a partial update
Currently when doing this test:

```c#
var indexUID = "BasicDocumentsUpdateTest";
Index index = await this.client.GetOrCreateIndex(indexUID);

// Add the documents
UpdateStatus update = await index.AddDocuments(new[]
{
    new Movie { Id = "1", Name = "Batman", Genre = "Action" },
    new Movie { Id = "2", Name = "Superman" }
});

// Update the documents
update = await index.UpdateDocuments(new[] { new Movie { Id = "1", Name = "Ironman" } });

// Check the documents have been updated and added
var docs = await index.GetDocuments<Movie>();
Assert.Equal("1", docs.First().Id);
Assert.Equal("Ironman", docs.First().Name);
Assert.Equal("Action", docs.First().Genre);
Assert.Equal("2", docs.ElementAt(1).Id);
Assert.Equal("Superman", docs.ElementAt(1).Name);
docs.ElementAt(1).Genre.Should().BeNull();
```

The `Assert.Equal("Action", docs.First().Genre);` line fails because `Genre` is `null` but it should not be. The [add-or-update-documents route](https://docs.meilisearch.com/references/documents.html#add-or-update-documents) should be a **partial update**, so the value a `Genre` should still be `"Action"`.

gethari:
Fixes #67

### DIFF
--- a/src/Meilisearch/Index.cs
+++ b/src/Meilisearch/Index.cs
@@ -120,7 +120,7 @@ namespace Meilisearch
             {
                 uri = QueryHelpers.AddQueryString(uri, new { primaryKey = primaryKey }.AsDictionary());
             }
-            var filteredDocuments = documents.ConvertToObjectWithoutPropertiesWithNullValues();
+            var filteredDocuments = documents.RemoveNullValues();
             responseMessage = await this.client.PutAsJsonAsync(uri, filteredDocuments);
 
             return await responseMessage.Content.ReadFromJsonAsync<UpdateStatus>();
--- a/src/Meilisearch/Index.cs
+++ b/src/Meilisearch/Index.cs
@@ -120,8 +120,8 @@ namespace Meilisearch
             {
                 uri = QueryHelpers.AddQueryString(uri, new { primaryKey = primaryKey }.AsDictionary());
             }
-
-            responseMessage = await this.client.PutAsJsonAsync(uri, documents);
+            var filteredDocuments = documents.ConvertToObjectWithoutPropertiesWithNullValues();
+            responseMessage = await this.client.PutAsJsonAsync(uri, filteredDocuments);
 
             return await responseMessage.Content.ReadFromJsonAsync<UpdateStatus>();
         }
--- a/src/Meilisearch/Index.cs
+++ b/src/Meilisearch/Index.cs
@@ -120,6 +120,7 @@ namespace Meilisearch
             {
                 uri = QueryHelpers.AddQueryString(uri, new { primaryKey = primaryKey }.AsDictionary());
             }
+
             var filteredDocuments = documents.RemoveNullValues();
             responseMessage = await this.client.PutAsJsonAsync(uri, filteredDocuments);
 
--- a/src/Meilisearch/ObjectExtensions.cs
+++ b/src/Meilisearch/ObjectExtensions.cs
@@ -28,7 +28,7 @@ namespace Meilisearch
         /// </summary>
         /// <param name="objectToTransform">Object to transform.</param>
         /// <returns>Returns the same IEnumerable with null keys removed. </returns>
-        public static object ConvertToObjectWithoutPropertiesWithNullValues<T>(this IEnumerable<T> objectToTransform)
+        public static object RemoveNullValues<T>(this IEnumerable<T> objectToTransform)
         {
             var result = objectToTransform.Select(item =>
             {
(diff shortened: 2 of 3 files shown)


src/Meilisearch/ObjectExtensions.cs
```diff
@@ -21,5 +22,26 @@ public static class ObjectExtensions
                 propInfo => char.ToLowerInvariant(propInfo.Name[0]) + propInfo.Name.Substring(1),
                 propInfo => propInfo.GetValue(source, null).ToString());
         }
+
+        /// <summary>
+        /// Iterates through Ienumerable and removes null keys.
+        /// </summary>
+        /// <param name="objectToTransform">Object to transform.</param>
+        /// <returns>Returns the same IEnumerable with null keys removed. </returns>
+        public static object ConvertToObjectWithoutPropertiesWithNullValues<T>(this IEnumerable<T> objectToTransform)
```
curquiza:
```suggestion
        public static object RemoveNullValues<T>(this IEnumerable<T> objectToTransform)
```

src/Meilisearch/Index.cs
```diff
@@ -120,8 +120,8 @@ public async Task<UpdateStatus> UpdateDocuments<T>(IEnumerable<T> documents, str
             {
                 uri = QueryHelpers.AddQueryString(uri, new { primaryKey = primaryKey }.AsDictionary());
             }
-
-            responseMessage = await this.client.PutAsJsonAsync(uri, documents);
+            var filteredDocuments = documents.ConvertToObjectWithoutPropertiesWithNullValues();
```
curquiza:
```suggestion
            var filteredDocuments = documents.RemoveNullValues();
```

curquiza (changes requested):
Hello @gethari, thanks for this PR. I suggest small naming changes, but everything is ok! Thanks for fixing this!!

src/Meilisearch/Index.cs
```diff
@@ -120,8 +120,8 @@ public async Task<UpdateStatus> UpdateDocuments<T>(IEnumerable<T> documents, str
             {
                 uri = QueryHelpers.AddQueryString(uri, new { primaryKey = primaryKey }.AsDictionary());
             }
-
-            responseMessage = await this.client.PutAsJsonAsync(uri, documents);
+            var filteredDocuments = documents.RemoveNullValues();
```
curquiza, line 124:
```suggestion

            var filteredDocuments = documents.RemoveNullValues();
```

src/Meilisearch/ObjectExtensions.cs
```diff
@@ -21,5 +22,26 @@ public static class ObjectExtensions
                 propInfo => char.ToLowerInvariant(propInfo.Name[0]) + propInfo.Name.Substring(1),
                 propInfo => propInfo.GetValue(source, null).ToString());
         }
+
+        /// <summary>
+        /// Iterates through Ienumerable and removes null keys.
+        /// </summary>
+        /// <param name="objectToTransform">Object to transform.</param>
+        /// <returns>Returns the same IEnumerable with null keys removed. </returns>
```
curquiza: Can you add the documentation for the typed parameter `T`? 🙂

curquiza (changes requested):

src/Meilisearch/ObjectExtensions.cs
gethari: Done @curquiza

curquiza: Thanks!

curquiza:
@gethari there is still this change to apply :) https://github.com/meilisearch/meilisearch-dotnet/pull/129/files#r646641400
You can just click on `Commit suggestion` button to accept my suggestion if you agree!

gethari:
> @gethari there is still this change to apply :) https://github.com/meilisearch/meilisearch-dotnet/pull/129/files#r646641400
> You can just click on `Commit suggestion` button to accept my suggestion if you agree!

I've already made that change. Not sure why comment is not shown as outdated. Here's how it looks in [Index.cs](https://github.com/gethari/meilisearch-dotnet/blob/403311fc7c13c099a09c20d91d5802d458402731/src/Meilisearch/Index.cs#L123) now 

image

curquiza:
The change I asked is only to add a new line after the bracket, you did not add it ;) 
That's why the comment is still here. You can accept my suggestion with the `commit suggestion` button

(this screenshot is an example)
commit_suggestion

curquiza (approved):
Merging!!
Thanks @gethari, it's really nice to have you as a contributor 😊 

bors merge

PR #130, opened 2021-06-03
Added delete index method in client

### LINKED ISSUE #121: Add client.DeleteIndex method
Currently, the only way to delete an index is from an `Index` instance like:

```c#
index.Delete();
```

We should be able to delete an `Index` from the client instance:

```c#
client.DeleteIndex("indexUID");
```

gethari:
Fixes #121

### DIFF
--- a/src/Meilisearch/MeilisearchClient.cs
+++ b/src/Meilisearch/MeilisearchClient.cs
@@ -3,6 +3,7 @@ namespace Meilisearch
     using System;
     using System.Collections.Generic;
     using System.Linq;
+    using System.Net;
     using System.Net.Http;
     using System.Net.Http.Json;
     using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


src/Meilisearch/MeilisearchClient.cs
```diff
@@ -189,5 +190,17 @@ public async Task<DumpStatus> GetDumpStatus(string uid)
...
+
+        /// <summary>
+        /// Deletes the index.
+        /// It's not a recovery delete. You will also lose the documents within the index.
+        /// </summary>
+        /// <param name="uid">unique dump identifier.</param>
+        /// <returns>Returns the status of delete operation.</returns>
+        public async Task<bool> DeleteIndex(string uid)
+        {
+            var responseMessage = await this.client.DeleteAsync($"/indexes/{uid}");
```
curquiza: To avoid code duplication, can you use the `Delete()` method in `Index` instead?

Like:
```c#
return await this.Index(uid).Delete();
```

curquiza (changes requested):
Thanks for your PR :)

FYI, but it does not impact your work on this PR: I noticed your commit was not recognized by Github as coming from your GitHub account. You did not enter your email as git config, so GitHub cannot link your commit with your GitHub account. That's too bad, because you will not appear in the contributor list :( maybe can you change your config and and push the next commits with this new configuration 🙂

gethari:
> Thanks for your PR :)
> 
> FYI, but it does not impact your work on this PR: I noticed your commit was not recognized by Github as coming from your GitHub account. You did not enter your email as git config, so GitHub cannot link your commit with your GitHub account. That's too bad, because you will not appear in the contributor list :( maybe can you change your config and and push the next commits with this new configuration 🙂

Oops my bad, I had my pc wiped some time back. Forgot to update the git config. Will do it and update the PR. Thanks for reviewing

curquiza:
bors try

curquiza (approved):
Thanks again @gethari!!!! 😄

curquiza:
bors merge

PR #142, opened 2021-08-09
Fix warning in test shorthand for nullable type

alallema:
Fix warning in tests:
`warning SA1125: Use shorthand for nullable types`


curquiza (approved):
thank you @alallema!!

PR #165, opened 2021-10-12
Add HttpRequest class to handle post and put requests and remove to the `charset-utf8` content type.

curquiza:
The goal is to implement our own HTTP methods (`PostAsJsonAsync`...) to remove the `charset-utf8` content type.

I did not find a smart way to override the methods in `HttpClient` or to extend `HttpClient`, I'm not a c# dev so I probably have missed something at the middle of all the errors, so here is my PR
- Adding a `HttpRequest` class that implements all the required method to communicate with the MeiliSearch server (`PostAsJsonAsync`...).
- Create an instance of `HttpReqest` in the `Client` class.

### DIFF
--- a/src/Meilisearch/HttpRequest.cs
+++ b/src/Meilisearch/HttpRequest.cs
@@ -1,0 +1,145 @@
+namespace Meilisearch
+{
+    using System;
+    using System.Net.Http;
+    using System.Net.Http.Json;
+    using System.Text;
+    using System.Text.Json;
+    using System.Threading.Tasks;
+
+    /// <summary>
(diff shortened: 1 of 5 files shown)


curquiza:
@alallema if you don't find any better implementation, we will merge this and I will open an issue to ask help

alallema (approved):
LGTM! 🎉 

After searching a bit I couldn't find a better way than yours.  Maybe it's because of the specific use we need maybe we missing something.

curquiza:
thank you @alallema for your investigation, i'm opening a new issue to find some help 😄

PR #173, opened 2021-10-14
 Add method to get the raw index information

### LINKED ISSUE #163: Add method to get the raw index information
⚠️ This issue is generated, it means the examples and the namings do not necessarily correspond to the language of this repository.
Also, if you are a maintainer, feel free to add any clarification and instruction about this issue.

Sorry if this is already partially/completely implemented, feel free to let me know about the state of this issue in the repo.

Related to https://github.com/meilisearch/integration-guides/issues/120

----

This SDK provides a method to get an index (`getIndex`) and this method returns an instance of `Index`. Some users want to get the raw version of the index, I mean the exact JSON response returned by the MeiliSearch server, not an instance.

Some SDKs already provide a way to get this raw index information but this is not convenient, example in PHP:

```php
$client->index('movies')->fetchRawInfo();
```

We need to add the `getRawIndex` method, accessible from the `Client` instance, to make it more convient.

Prototype:

```
getRawIndex(string indexUID) -> object/struct/dict of the MeiliSearch response
```

Ex in PHP:
```php
$client->getRawIndex('movies');
```

TODO:
- [ ] Add `getRawIndex()` method to the `Client` instance without duplicating the code (if `fetchRawInfo()` already exists for example)
- [ ] Add tests

alejandrocq:
PR for issue #163 

I have added the method `MeilisearchClient#getRawIndex` (which returns `JsonElement`, not sure if that fits your needs) and the test `MeilisearchClientTests#getRawIndex`. Let me know if it is ok :)

### DIFF
--- a/src/Meilisearch/MeilisearchClient.cs
+++ b/src/Meilisearch/MeilisearchClient.cs
@@ -5,6 +5,7 @@ namespace Meilisearch
     using System.Linq;
     using System.Net.Http;
     using System.Net.Http.Json;
+    using System.Text.Json;
     using System.Threading.Tasks;
     using Meilisearch.Extensions;
 
@@ -100,6 +101,17 @@ namespace Meilisearch
         public async Task<Index> GetIndex(string uid)
         {
             return await this.Index(uid).FetchInfo();
+        }
+
+        /// <summary>
+        /// Gets an index in raw format.
+        /// </summary>
+        /// <param name="uid">UID of the index to get.</param>
+        /// <returns>A <see cref="JsonElement"/> which represents the raw index as a JSON object.</returns>
+        public async Task<JsonElement> GetRawIndex(string uid)
+        {
+            var json = await (await Meilisearch.Index.GetRaw(this.http, uid)).Content.ReadAsStringAsync();
+            return JsonDocument.Parse(json).RootElement;
         }
 
         /// <summary>
--- a/tests/Meilisearch.Tests/IndexTests.cs
+++ b/tests/Meilisearch.Tests/IndexTests.cs
@@ -4,15 +4,15 @@ namespace Meilisearch.Tests
     using System.Linq;
     using System.Threading.Tasks;
     using FluentAssertions;
+    using HttpClientFactoryLite;
     using Xunit;
 
     [Collection("Sequential")]
     public class IndexTests : IAsyncLifetime
     {
-        private MeilisearchClient defaultClient;
-        private string defaultPrimaryKey;
-
-        private IndexFixture fixture;
+        private readonly MeilisearchClient defaultClient;
+        private readonly string defaultPrimaryKey;
+        private readonly IndexFixture fixture;
 
         public IndexTests(IndexFixture fixture)
         {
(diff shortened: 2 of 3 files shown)


curquiza (changes requested):
Can you put your test in IndexTests.cs? :)

alejandrocq:
@curquiza Sure! Let me get some time, and I will change it.

src/Meilisearch/MeilisearchClient.cs
```diff
@@ -98,6 +99,17 @@ public async Task<Index> GetIndex(string uid)
...
         }
 
+        /// <summary>
+        /// Gets an index in raw format.
+        /// </summary>
+        /// <param name="uid">UID of the index to get.</param>
+        /// <returns>A <see cref="JsonElement"/> which represents the raw index as a JSON object.</returns>
+        public async Task<JsonElement> GetRawIndex(string uid)
+        {
+            var json = await (await this.http.GetAsync($"indexes/{uid}")).Content.ReadAsStringAsync();
```
curquiza: `FetchInfo()` method in Index class already does this API call:

https://github.com/meilisearch/meilisearch-dotnet/blob/ad5ea3013528ca0b6a0ea59caba13af6de18b660/src/Meilisearch/Index.cs#L49

Can you change to avoid the code base having `await this.http.GetAsync($"indexes/{uid}")` twice?
For example, create a `GetRaw` method in `Index` class that would be used by `FetchInfo()` and `GetRawInfex()` :)

curquiza (changes requested):
Your test is still in the `tests/Meilisearch.Tests/MeilisearchClientTests.cs` instead of `meilisearch-dotnet/tests/Meilisearch.Tests/IndexTests.cs`

src/Meilisearch/MeilisearchClient.cs
alejandrocq: Sure! Good point. Working on it right now.

alejandrocq: Done! Let me know your thoughts.

tests/Meilisearch.Tests/IndexTests.cs
```diff
@@ -207,5 +207,18 @@ public async Task WhenIndexNoLongerExists_DeleteIfExists_ShouldReturnFalse()
...
+        public async Task GetRawIndex()
+        {
+            await this.fixture.SetUpBasicIndex("BasicIndex");
+            var httpClient = ClientFactory.Instance.CreateClient<MeilisearchClient>();
+            MeilisearchClient ms = new MeilisearchClient(httpClient);
+
+            var rawIndex = await ms.GetRawIndex("BasicIndex");
+
+            rawIndex.GetProperty("uid").GetString().Should().Be("BasicIndex");
+            rawIndex.GetProperty("name").GetString().Should().Be("BasicIndex");
```
curquiza:
```suggestion
```

curquiza (approved):
Thanks @alejandrocq 

bors merge

If you are participating in [Hacktoberfest](https://blog.meilisearch.com/contribute-hacktoberfest-2021/), and you would like to receive a small gift from MeiliSearch too, please complete [this form](https://meilisearch.typeform.com/to/rPexR5gq).

PR #172, opened 2021-10-14
Refactor http client

Vivek-abstract:
Refactoring as per issue:
https://github.com/meilisearch/meilisearch-dotnet/issues/167

I created a class for extension methods which will handle the custom extension methods like PostAsJsonAsync. Now we can use HttpClient directly and make requests.
I wasn't able to run the tests locally due to docker issues with hyper v. Let me know if any tests fail or if any other change is required. 

This is my first open source PR for .NET so let me know your feedback 😄

PR #178, opened 2021-10-17
Add method to fetch list of raw indexes

### LINKED ISSUE #155: Method to get all the indexes should return a list of instance
Currently, the `get_all_indexes` method returns a raw list of object, but not a list of `Index` instance. This is the only index-related method that does not return an `Index` instance.

An instance is more convenient to manipulate than a raw object. See this [example](https://github.com/meilisearch/meilisearch-python/issues/258#issuecomment-835866135).

TODO

- [x] Change the return of the current `get_all_indexes`: make it return a list of `Index` instances
- [ ] Provide `get_all_raw_indexes`: a method that returns the list of object of indexes, so the response of the MeiliSearch server.
- [ ] Add tests

⚠️ No code duplication: use `get_all_raw_indexes` in `get_all_indexes`.

Related to https://github.com/meilisearch/integration-guides/issues/122

⚠️ Also, this issue is generated, and `get_all_indexes` might be named differently in this package. Keep the already existing way of naming in this package to stay idomatic with the language and the repository.

Sorry if this is already partially/completely implemented, feel free to let me know about the state of this issue.

Vivek-abstract:
Related to issue:
https://github.com/meilisearch/meilisearch-dotnet/issues/155

Added a method GetAllRawIndexes which will return the response from the server as a enumerable of dictionary of key value pairs.
I didn't modify GetAllIndexes to use this new method because I think its best to keep it simple, adding conversion logic from dictionary to index would make it complicated. Let me know what you think

P.S. the 2 warnings would be fixed by the other PR https://github.com/meilisearch/meilisearch-dotnet/pull/172


curquiza (changes requested):
I would not add a new dependency. Can we just make it work like this PR (https://github.com/meilisearch/meilisearch-dotnet/pull/173) suggests, without any new dependencies? :)

Thanks for your PR :)

Vivek-abstract:
@curquiza I changed it to use existing packages and return an ienumerable of JsonElements instead. 👍

src/Meilisearch/MeilisearchClient.cs
```diff
@@ -92,6 +92,27 @@ public async Task<Index> UpdateIndex(string uid, string primarykeytoChange)
             return await this.Index(uid).Update(primarykeytoChange);
         }
 
+        /// <summary>
+        /// Gets all the raw indexes for the instance as returned by the resposne of the Meilisearch server. Throws error if the index does not exist.
+        /// </summary>
+        /// <returns>An IEnumerable of indexes in JsonElement format.
+        /// Ex: [{"uid": "movies", "name":"movies", "createdAt":"2021-10-17T08:24:15.222102668Z",updatedAt:"2021-10-17T08:24:15.222102668Z",primaryKey:"movieId"}]. </returns>
```
curquiza:
```suggestion
        /// <returns>An IEnumerable of indexes in JsonElement format.</returns>
```

curquiza (approved):
Thank you again @Vivek-abstract 

bors merge

If you are participating in [Hacktoberfest](https://blog.meilisearch.com/contribute-hacktoberfest-2021/), and you would like to receive a small gift from MeiliSearch too, please complete [this form](https://meilisearch.typeform.com/to/rPexR5gq).

PR #197, opened 2021-11-23
Remove dependancy on Microsoft.AspNetCore.WebUtilities

### LINKED ISSUE #176: Remove Microsoft.AspNetCore.WebUtilities to increase the surface area of the project
Hello!

I was having a peek at the dependencies of this project, and I encountered the `Microsoft.AspNetCore.WebUtilities` dependency. This dependency has itself 2 dependencies on other projects..

What is the reason for this package? As far as I can see, it is only used to generate some query strings:
https://github.com/meilisearch/meilisearch-dotnet/blob/8d5b14cc2a12a407b88ee4f9b41c0b80bf6c6548/src/Meilisearch/Index.cs#L129

I believe these do not belong in this library for several reasons:

- I see this library as general purpose; any type of project should be able to use it. So why do we have an `AspNetCore` package dependency? 
- Why introduce a dependency to generate some `?query=strings&and=stuff`? :) This can easily be done with normal C#.

Removing this dependency will increase the surface area that cna be reached with this library because it needs fewer dependencies to function and could increase adoption by removing a "wonky" dependency :)

juchom:
# Pull Request

## What does this PR do?
Fixes #176 

This method is actually used in order to work with `QueryHelpers.AddQueryString`

```C#
public static IDictionary<string, string> AsDictionary(this object source, BindingFlags bindingAttr = BindingFlags.DeclaredOnly | BindingFlags.Public | BindingFlags.Instance)
{
    return source.GetType().GetProperties(bindingAttr).Where(p => p.GetValue(source, null) != null).ToDictionary(
        propInfo => char.ToLowerInvariant(propInfo.Name[0]) + propInfo.Name.Substring(1),
        propInfo => propInfo.GetValue(source, null).ToString());
}
```

I've created a new one that takes an object and transform it to a query string.

```C#
public static string ToQueryString(this object source, BindingFlags bindingAttr = BindingFlags.DeclaredOnly | BindingFlags.Public | BindingFlags.Instance)
{
    var values = source.GetType().GetProperties(bindingAttr)
    .Where(p => p.GetValue(source, null) != null)
    .Select(p =>
        Uri.EscapeDataString(char.ToLowerInvariant(p.Name[0]) + p.Name.Substring(1)) + "=" + Uri.EscapeDataString(p.GetValue(source, null).ToString()));
    var queryString = string.Join("&", values);
    return queryString;
}
```

I've change in `Index.cs`

```C#
// FROM
// TO
uri = $"{uri}?{new { primaryKey = primaryKey }.ToQueryString()}";
```

The query string is appended by hand, you may want to have a helper method to do this

Some tests have been added to make sure queries are the same.

### DIFF
--- a/src/Meilisearch/ObjectExtensions.cs
+++ b/src/Meilisearch/ObjectExtensions.cs
@@ -25,7 +25,7 @@ namespace Meilisearch
         }
 
         /// <summary>
-        /// Transforms an MeiliSearch object into an url encoded query string.
+        /// Transforms a MeiliSearch object into an URL encoded query string.
         /// </summary>
         /// <param name="source">Object to transform.</param>
         /// <param name="bindingAttr">Binding flags.</param>
--- a/src/Meilisearch/ObjectExtensions.cs
+++ b/src/Meilisearch/ObjectExtensions.cs
@@ -1,5 +1,6 @@
 namespace Meilisearch
 {
+    using System;
     using System.Collections.Generic;
     using System.Dynamic;
     using System.Linq;
@@ -21,6 +22,22 @@ namespace Meilisearch
             return source.GetType().GetProperties(bindingAttr).Where(p => p.GetValue(source, null) != null).ToDictionary(
                 propInfo => char.ToLowerInvariant(propInfo.Name[0]) + propInfo.Name.Substring(1),
                 propInfo => propInfo.GetValue(source, null).ToString());
+        }
+
+        /// <summary>
+        /// Transforms an MeiliSearch object into an url encoded query string.
+        /// </summary>
+        /// <param name="source">Object to transform.</param>
+        /// <param name="bindingAttr">Binding flags.</param>
+        /// <returns>Returns an url encoded query string.</returns>
+        public static string ToQueryString(this object source, BindingFlags bindingAttr = BindingFlags.DeclaredOnly | BindingFlags.Public | BindingFlags.Instance)
+        {
+            var values = source.GetType().GetProperties(bindingAttr)
+            .Where(p => p.GetValue(source, null) != null)
+            .Select(p =>
+                Uri.EscapeDataString(char.ToLowerInvariant(p.Name[0]) + p.Name.Substring(1)) + "=" + Uri.EscapeDataString(p.GetValue(source, null).ToString()));
+            var queryString = string.Join("&", values);
+            return queryString;
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


sander1095:
Resolves #176

alallema:
Hi @juchom,
Thank you so much for contributing to Meilisearch ❤️!
@curquiza is on holiday and will review your PR once they are back.

juchom:
I've ran some benchmarks also

```C#
[MemoryDiagnoser]
public class BenchmarkTest
{
	[Benchmark(Baseline = true)]
	[Arguments("/api/url", "key")]
	public string QueryAspNetCore(string uri, string primaryKey)
	{
		return QueryHelpers.AddQueryString(uri, new { primaryKey = primaryKey }.AsDictionary());
	}

	[Benchmark]
	[Arguments("/api/url", "key")]
	public string QueryCustom(string uri, string primaryKey)
	{
 		return $"{uri}?{new { primaryKey = primaryKey }.ToQueryString()}";
	}

	[Benchmark]  // Not used in project
	[Arguments("/api/url", "key")]
	public string QueryCustomSb(string uri, string primaryKey)
	{
		var sb = new StringBuilder();
		return sb.Append(uri).Append("?").Append(new { primaryKey = primaryKey }.ToQueryString()).ToString();
	}
}
```

```
|          Method |      uri | primaryKey |     Mean |    Error |   StdDev | Ratio | RatioSD |  Gen 0 | Allocated |
|---------------- |--------- |----------- |---------:|---------:|---------:|------:|--------:|-------:|----------:|
| QueryAspNetCore | /api/url |        key | 705.2 ns | 14.01 ns | 35.91 ns |  1.00 |    0.00 | 0.2928 |     920 B |
|     QueryCustom | /api/url |        key | 479.8 ns |  9.52 ns | 19.67 ns |  0.68 |    0.05 | 0.1760 |     552 B |
|   QueryCustomSb | /api/url |        key | 551.7 ns | 11.05 ns | 27.10 ns |  0.78 |    0.05 | 0.2422 |     760 B |
```

sander1095:
I can't say anything about the quality of the benchmark, I don't have any experience with it. But if that is reliable, it looks good! Well done!

src/Meilisearch/ObjectExtensions.cs
```diff
@@ -23,6 +24,22 @@ public static class ObjectExtensions
                 propInfo => propInfo.GetValue(source, null).ToString());
         }
 
+        /// <summary>
+        /// Transforms an MeiliSearch object into an url encoded query string.
```
curquiza:
```suggestion
        /// Transforms a MeiliSearch object into an URL encoded query string.
```

curquiza (changes requested):

curquiza (approved):
Thanks @juchom for this quality PR and the explanation/details!

bors merge

PR #207, opened 2022-01-03
Rename all async methods with an Async suffix

### LINKED ISSUE #181: Use ConfigureAwait(false) to improve performance of the library
Please read:

https://github.com/Microsoft/vs-threading/blob/main/doc/cookbook_vs.md#should-i-await-a-task-with-configureawaitfalse
https://medium.com/bynder-tech/c-why-you-should-use-configureawait-false-in-your-library-code-d7837dce3d7f

Not using `ConfigureAwait(false)` in a library that might be used by an app that has a SynchronizationContext could result in a loss of performance and perhaps some issues.

This library should use `ConfigureAwait(false)` everywhere to prevent this

### LINKED ISSUE #196: Suffix async methods with Async
Today .NET ecosystem suffix all asynchronous methods with Async.

Since the client .net client is currently all async it would be good to suffix this methods with Async.

This would leave space to create a sync api if needed for some users.

Also, passing CancellationToken and propagating it to httpclient would be a good idea.

juchom:
# Pull Request

## What does this PR do?
Fixes #196, #181

Renamed all async methods with an `Async` suffix to match .net coding style.
All async methods now contains an optional `CancellationToken` parameter.
All async calls are using `ConfigureAwait(false)`

## PR checklist

Another pair of eyes could be necessary to check nothing is missing !

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -18,7 +18,7 @@ delete_documents_1: |-
 search_post_1: |-
 get_update_1: |-
 get_all_updates_1: |-
-  var updates = await client.Index("movies").GetAllUpdateStatus();
+  var updates = await client.IndexAsync("movies").GetAllUpdateStatus();
   foreach (UpdateStatus update in updates)
   {
       Console.WriteLine(update.Status);
@@ -37,25 +37,25 @@ get_ranking_rules_1: |-
 update_ranking_rules_1: |-
 reset_ranking_rules_1: |-
 get_distinct_attribute_1: |-
-  string result = await client.Index("shoes").GetDistinctAttribute();
+  string result = await client.IndexAsync("shoes").GetDistinctAttribute();
 update_distinct_attribute_1: |-
-  UpdateStatus result = await client.Index("shoes").UpdateDistinctAttribute("skuid");
+  UpdateStatus result = await client.IndexAsync("shoes").UpdateDistinctAttribute("skuid");
 reset_distinct_attribute_1: |-
-  UpdateStatus result = await client.Index("shoes").ResetDistinctAttribute();
+  UpdateStatus result = await client.IndexAsync("shoes").ResetDistinctAttribute();
 get_searchable_attributes_1: |-
 update_searchable_attributes_1: |-
 reset_searchable_attributes_1: |-
 get_filterable_attributes_1: |-
-  IEnumerable<string> attributes = await client.Index("movies").GetFilterableAttributes();
+  IEnumerable<string> attributes = await client.IndexAsync("movies").GetFilterableAttributes();
   foreach (string attribute in attributes)
   {
       Console.WriteLine(attribute);
   }
 update_filterable_attributes_1: |-
   List<string> attributes = new() { "genres", "director" };
-  UpdateStatus result = await client.Index("movies").UpdateFilterableAttributes(attributes);
+  UpdateStatus result = await client.Index("movies").UpdateFilterableAttributesAsync(attributes);
 reset_filterable_attributes_1: |-
-  UpdateStatus result = await client.Index("movies").ResetFilterableAttributes();
+  UpdateStatus result = await client.Index("movies").ResetFilterableAttributesAsync();
 get_displayed_attributes_1: |-
 update_displayed_attributes_1: |-
 reset_displayed_attributes_1: |-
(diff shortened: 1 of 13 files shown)


src/Meilisearch/Index.cs
```diff
@@ -158,8 +168,8 @@ public async Task<UpdateStatus> AddDocuments<T>(IEnumerable<T> documents, string
                 uri = $"{uri}?{new { primaryKey = primaryKey }.ToQueryString()}";
             }
 
-            responseMessage = await this.http.PostJsonCustomAsync(uri, documents);
-            return await responseMessage.Content.ReadFromJsonAsync<UpdateStatus>();
+            responseMessage = await this.http.PostJsonCustomAsync(uri, documents, cancellationToken: cancellationToken).ConfigureAwait(false);
```
curquiza, line 171: Why should we use `ConfigureAwait(false)` here again since this method is already called in `PostJsonCustomAsync`?
Sorry I'm not a dotnet expert, I might have missed something

src/Meilisearch/Extensions/HttpExtensions.cs
```diff
@@ -16,13 +17,14 @@ public static class HttpExtensions
         /// <param name="client">HttpClient.</param>
         /// <param name="uri">Endpoint.</param>
         /// <param name="body">Body sent.</param>
+        /// <param name="cancellationToken">The cancellation token for this call.</param>
         /// <typeparam name="T">Type of the body to send.</typeparam>
         /// <returns>Returns the HTTP response from the MeiliSearch server.</returns>
-        public static async Task<HttpResponseMessage> PostJsonCustomAsync<T>(this HttpClient client, string uri, T body)
+        public static async Task<HttpResponseMessage> PostJsonCustomAsync<T>(this HttpClient client, string uri, T body, CancellationToken cancellationToken = default)
```
curquiza, line 23: Can you rename the `cancellationToken` variable into `ct` to make it less verbose? :)

curquiza (changes requested):
Thanks a lot for this PR @juchom!!

src/Meilisearch/Index.cs
juchom: We have two awaited calls the outer one `this.http.PostJsonCustomAsync` and the inner one inside `PostJsonCustomAsync`.

Each time we do `async` / `await` we create a state machine, `ConfigureAwait(false)` is just here to say when the awaited call is done it can use any available thread in the thread pool instead of waiting for the original thread.

We call it for both just to avoid the extra cost of waiting for the original thread to become available.

src/Meilisearch/Extensions/HttpExtensions.cs
juchom: I would not do it for two reasons :

`cancellationToken` is the name used by almost every projects (Dapper, MediatR, XUnit,...)

Since `cancellationToken` is an optional parameter, if you have to specify it, I think it's better to have a clear name.

With `ct`, it may be a cancellationToken or something else
```csharp
http.PostJsonCustomAsync<T>(uri: "https://domain.tld", 
                            body: obj,
                            ct: ct);
```

With `cancellationToken`, it's clear
```csharp
http.PostJsonCustomAsync<T>(uri: "https://domain.tld", 
                            body: obj,
                            cancellationToken: ct);
```

I'm not sure of the rule, but it says something like "you write the code once, you read it ten times".

Just let me know what you prefer and I'll fix it.

curquiza: You're right
I mainly would to avoid `cancellationToken: cancellationToken` being written too many times since it's "heavy", but it might be less clear

src/Meilisearch/Index.cs
curquiza: Ok! Thanks for the explanation!

curquiza (changes requested):
Can you update the code-sample file as well? 🙂
https://github.com/meilisearch/meilisearch-dotnet/blob/main/.code-samples.meilisearch.yaml

curquiza:
@sander1095, if you are around, feel free to check your issue (#207) is implemented as you expected :)

.code-samples.meilisearch.yaml
```diff
@@ -18,7 +18,7 @@ delete_documents_1: |-
 search_post_1: |-
 get_update_1: |-
 get_all_updates_1: |-
-  var updates = await client.Index("movies").GetAllUpdateStatus();
+  var updates = await client.IndexAsync("movies").GetAllUpdateStatus();
```
curquiza: `IndexAdync` does not exist
`GetAllUpdateStatus` should be `GetAllUpdateStatusAsync`

curquiza (changes requested):

sander1095:
> @sander1095, if you are around, feel free to check your issue (#207) is implemented as you expected :)

I had a quick look, LGTM! 👍

.code-samples.meilisearch.yaml
juchom: Sorry, I was sure I had fixed these lines. Should be good now.

curquiza (approved):
Perfect! Thank you so much @juchom for this PR and @sander1095 for the review!

bors merge

Issue #182, opened 2021-10-19
NDJSON/CSV methods to add and update documents


curquiza:
⚠️ This issue is generated, it means the nameing might be done differently in this package (ex: `add_documents_json` instead of `addDocumentsJson`). Keep the already existing way of naming in this package to stay idiomatic with the language and this repository.

📣 We strongly recommend doing multiple PRs to solve all the points of this issue

[MeiliSearch v0.23.0](https://github.com/meilisearch/MeiliSearch/releases/tag/v0.23.1) introduces two changes:
- new valid formats to push data files, additionally to the JSON format: CSV and NDJSON formats.
- it enforces the `Content-type` header for every route requiring a payload (`POST` and `PUT` routes)

---

Here are the expected changes to completely close the issue:

- [x] Currently, the SDKs always send `Content-Type: application/json` to every request. Only the `POST` and `PUT` requests should send the `Content-Type: application/json` and not the `DELETE` and `GET` ones.

- [x] Add the following methods and 🔥 **the associated tests** 🔥 to ADD the documents. Depending on the format type (`csv` or `ndjson`) the SDK should send `Content-Type: application/x-dnjson` or `Content-Type: text/csv`)
  - [x] `addDocumentsJson(string docs, string primaryKey)`
  - [x] `addDocumentsCsv(string docs, string primaryKey)`
  - [x] `addDocumentsCsvInBatches(string docs, int batchSize, string primaryKey)`
  - [x] `addDocumentsNdjson(string docs, string primaryKey)`
  - [x] `addDocumentsNdjsonInBatches(string docs, int batchSize, string primaryKey)`

- [x] Add the following methods and 🔥 **the associated tests** 🔥 to UPDATE the documents. Depending on the format type (`csv` or `ndjson`) the SDK should send `Content-Type: application/x-dnjson` or `Content-Type: text/csv`)
  - [x] `updateDocumentsJson(string docs, string primaryKey)`
  - [x] `updateDocumentsCsv(string docs, string primaryKey)`
  - [x] `updateDocumentsCsvInBatches(string docs, int batchSize, string primaryKey)`
  - [x] `updateDocumentsNdjson(string docs, string primaryKey)`
  - [x] `updateDocumentsNdjsonInBatches(string docs, int batchSize, string primaryKey)`

`docs` are the documents sent as `String`
`primaryKey` is [the primary key of the index](https://docs.meilisearch.com/learn/core_concepts/indexes.html#primary-key)
[...]

PR #213, opened 2022-01-10
Refactor code to prepare support for NDJSON and CSV

juchom:
# Pull Request
The main idea of this pull request is to cleanup and refactor some code in order to easily add support for CSV and NDJSON.

The only format supported is still JSON. I have renamed the methods according to the issue.

Batches methods were a bit akward and difficult to read / understand, so I removed the `BatchOperationAsync` method and added an `EnumerableExtensions` class which gives the `Chunk`.

Documents Add / Update unit tests are now `Theory` so we can easily extends them for CSV and NDJSON.

Documentation has been updated.

@curquiza let me know if you need help to review the code.

## What does this PR do?
First step for #182

## PR checklist

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -150,7 +150,7 @@ getting_started_add_documents_md: |-
               var movies = JsonSerializer.Deserialize<IEnumerable<Movie>>(jsonString, options);
 
               var index = client.Index("movies");
-              await index.AddDocumentsAsync<Movie>(movies);
+              await index.AddDocumentsJsonAsync<Movie>(movies);
           }
       }
   }
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ riderModule.iml
 /.idea/
 .vscode
 /.vs
+*.user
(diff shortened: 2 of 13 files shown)


juchom:
I was digging a bit further to add NDJSON and CSV support and I found something that needs to be discussed a bit.

I think there is an issue with the JsonSerializerOptions here :

https://github.com/meilisearch/meilisearch-dotnet/blob/8d63ed44b0eabebcc9cd4fbc3a59dfc5831391f7/src/Meilisearch/MeilisearchClient.cs#L22-L26

And this code

https://github.com/meilisearch/meilisearch-dotnet/blob/8d63ed44b0eabebcc9cd4fbc3a59dfc5831391f7/src/Meilisearch/ObjectExtensions.cs#L49-L63

Used here
https://github.com/meilisearch/meilisearch-dotnet/blob/8d63ed44b0eabebcc9cd4fbc3a59dfc5831391f7/src/Meilisearch/Index.cs#L212

First issue is the custom function `RemoveNullValues<T>` is doing the same as the JsonSerializerOptions `DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull`, duplicate code that can be removed easy one.

Second issue which is more serious, removing null properties is a bad decision.

Here is why :

Imagine we have this object we want to add to Meilisearch

```csharp
public class UserAccount
{
    public int Id { get; set; }
    public bool? HasAcceptedLicense { get; set; }
}
```

When we add the object the first time we have the following values

```
Id = 1
Name = John
HasAcceptedLicense = true
```

A few days later, you change the license and require the user to accept or refuse it and you want to set `HasAcceptedLicense` to `null` which means "the user has not said if he accepts or refuses the new license".

You won't be able to do that with the current `JsonSerializerOptions`.

Last important thing is `CSV` is not as flexible as `JSON` when serializing, there is no optional field, the schema is the same for all records.

With the following records and actual options

```
Id = 1
Name = John
HasAcceptedLicense = true

Id = 2
Name = Jane
HasAcceptedLicense = null
```

The `Jane` record would not send the HasAcceptedLicense to Meilisearch with `JSON` or `NDJSON`, but it `CSV` it would send it.

To conclude 

The best option would be to remove the custom `RemoveNullValues<T>` and `DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull` this would avoid the issues above.

If you're ok with that, I can add a small commit to this PR with the changes.

juchom:
Before working on this PR, I suggest that we first finish the work to support Meilisearch 0.25.

Then, I will take some changes from this big PR and create smaller ones to prepare the support for NDJSON and CSV. It will be easier to review.

curquiza:
> Before working on this PR, I suggest that we first finish the work to support Meilisearch 0.25.

Completely agreed!

PR #214, opened 2022-01-13
Update client to support Meilisearch 0.25

juchom:
## What does this PR do?
This pull request makes sure all tests are working with the Meilisearch 0.25.

I've removed the same methods and tests from your PR.

I've renamed `UpdateStatus` to `TaskInfo` and not `TaskStatus` because `TaskStatus` is already used by the framework.

Since Index are now created asynchronous, methods are now returning `TaskStatus` and we have to await for it to finish and then fetch the index.

```csharp
var indexUid = "IndexFetchExistingIndexPrimaryKeyTest";
var taskInfo = await this.defaultClient.CreateIndexAsync(indexUid, this.defaultPrimaryKey);
await this.defaultClient.Tasks.WaitForPendingTaskAsync(taskInfo.Uid);
var createIndex = await this.defaultClient.GetIndexAsync(indexUid);
```

It's not very user friendly and I think we should hide from the end user this complexity and make.

The same code should be like this (the way, it is working right now)

```csharp
var indexUid = "IndexFetchExistingIndexPrimaryKeyTest";
// We await in this call the end of the create index task
var createIndex = await this.defaultClient.CreateIndexAsync(indexUid, this.defaultPrimaryKey);
```

Also `Task` related functions and `Dump` related functions are in their own class.

This reduce a bit the size of the `Index.cs` class


curquiza:
Hello @juchom thanks a lot for your involvement in this project.
Making this package up to date with MeiliSearch v0.25.0 is a huge work and it's not that easy when you don't know perfectly the changelog. I would rather warn you since I don't want to make you lose your time .
As you might have seen, I've already started the work on this branch (https://github.com/meilisearch/meilisearch-dotnet/pull/202). I have to do the rest of it and merge it into this PR before merging on `main`.
Thanks for your work anyway, it will definitely help me update this package. 

Sorry for the delay in making this package compatible with MeiliSearch v0.25.0, but we are a small team, and the changes in v0.25.0 are huge. We succeeded to release some clients by the release deadline, but not all of them.

Thanks again for being around 😁

PR #215, opened 2022-01-16
Changes regarding new task API

curquiza:
Breaking changes

- `CreateIndex()`, `UpdateIndex()` and `DeleteIndex()` are now asynchrone regarding MeiliSearch and do not return a `Index` instance anymore, but a task object. Please use `Index()` method instead.
- Rename `index.GetAllUpdateStatus()` into `index.GetTasks()`
- Rename `index.getUpdateStatus()` into `index.GetTask()`
- `waitForPendingUpdate` is renamed into `waitForTask` and is accessible from `Index` and from `Client`
- Rename the `UpdateStatus` class into `TaskInfo` since `TaskStatus` and `Task` are reserved words in C#.

Improvements

- Add new methods `client.GetTasks()` and  `Client.GetTask()`

⚠️ Final code sample changes coming in another PR


README.md
```diff
@@ -220,11 +220,11 @@ var index = await client.GetIndexAsync("movies");
 #### Add or Update Documents <!-- omit in toc -->
 
 ```c#
-var updateStatus = await index.AddDocumentsAsync(new Movie[] { new Movie { Id = "1", Title = "Carol" } } );
-var updateStatus = await index.UpdateDocumentsAsync(new Movie[] { new Movie { Id = "1", Title = "Carol" } } );
+var task = await index.AddDocumentsAsync(new Movie[] { new Movie { Id = "1", Title = "Carol" } } );
```
brunoocasali, line 223: I was reading about naming conventions in C# and I found this really interesting guide about using `var`'s, in this case, we could explicitly set that the response is a `TaskInfo` instead of var, what do you think? 🥼 


https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions#implicitly-typed-local-variables

brunoocasali: Actually not just here, but in the other places where the type is not clear :)

src/Meilisearch/MeilisearchClient.cs
```diff
@@ -26,6 +26,7 @@ public class MeilisearchClient
         };
 
         private readonly HttpClient http;
+        private TaskEndpoint taskEndpoint;
```
brunoocasali: Is a common practice to prefix private fields with `_`.

https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions#camel-case

src/Meilisearch/MeilisearchClient.cs
```diff
@@ -232,15 +284,18 @@ public async Task<DumpStatus> GetDumpStatusAsync(string uid, CancellationToken c
...
-        /// Deletes the index.
-        /// It's not a recovery delete. You will also lose the documents within the index.
+        /// Create a local reference to a task, without doing an HTTP call.
         /// </summary>
-        /// <param name="uid">unique dump identifier.</param>
-        /// <param name="cancellationToken">The cancellation token for this call.</param>
-        /// <returns>Returns the status of delete operation.</returns>
-        public async Task<bool> DeleteIndexAsync(string uid, CancellationToken cancellationToken = default)
+        /// <returns>Returns a Task instance.</returns>
+        private TaskEndpoint TaskEndpoint()
```
brunoocasali, line 290: Is there any particular reason to not instantiate here? https://github.com/meilisearch/meilisearch-dotnet/pull/215/files#diff-0fb72721862f4f133d8d58e4fcb74332c6d5f72bc4652a6e3b67309376296bceR41

src/Meilisearch/TaskEndpoint.cs
```diff
@@ -0,0 +1,107 @@
...
+        private HttpClient http;
+
+        /// <summary>
+        /// Gets the tasks.
+        /// </summary>
+        /// <param name="cancellationToken">The cancellation token for this call.</param>
+        /// <returns>Returns a list of the tasks.</returns>
+        public async Task<Result<IEnumerable<TaskInfo>>> GetTasksAsync(CancellationToken cancellationToken = default)
+        {
+            return await this.http.GetFromJsonAsync<Result<IEnumerable<TaskInfo>>>($"/tasks", cancellationToken: cancellationToken)
```
brunoocasali: I think we do need the interpolation here :)

```suggestion
            return await this.http.GetFromJsonAsync<Result<IEnumerable<TaskInfo>>>("/tasks", cancellationToken: cancellationToken)
```

src/Meilisearch/TaskEndpoint.cs
```diff
@@ -0,0 +1,107 @@
...
+            int taskUid,
+            double timeoutMs = 5000.0,
+            int intervalMs = 50,
+            CancellationToken cancellationToken = default)
+        {
+            DateTime endingTime = DateTime.Now.AddMilliseconds(timeoutMs);
+
+            while (DateTime.Now < endingTime)
+            {
+                var task = await this.GetTaskAsync(taskUid, cancellationToken).ConfigureAwait(false);
```
brunoocasali, line 82: I think that idea about the `var` remains here too!

src/Meilisearch/TaskEndpoint.cs
```diff
@@ -0,0 +1,107 @@
...
+
+                if (task.Status != "enqueued" && task.Status != "processing")
```
brunoocasali, line 84: Am I too annoying about this idea of adding these as a constant? I've bothered @alallema with that question in the Java PR as well...

src/Meilisearch/TaskInfo.cs
```diff
@@ -0,0 +1,35 @@
...
+
+    /// <summary>
+    /// Update Status of the actions done.
+    /// </summary>
+    public class TaskInfo
+    {
+        /// <summary>
+        /// Gets or sets Uid for the task.
+        /// </summary>
+        public int Uid { get; set; }
```
brunoocasali, line 13: This comment could be applied to all of the properties. 

These `get; set;` are always required? So there are some of these properties that would not be necessary for example, only `Status` requires `set` but not `Uid`, `Type`. The general idea is to only use `get` and `set` when is really needed.

tests/Meilisearch.Tests/IndexTests.cs
```diff
@@ -47,75 +50,89 @@ public async Task IndexCreationWithPrimaryKey()
...
         {
             var indexUid = "BasicUsageOfIndexMethodTest";
-            var index = this.defaultClient.Index(indexUid);
+            var index = this.client.Index(indexUid);
             index.Uid.Should().Be(indexUid);
             index.PrimaryKey.Should().BeNull();
-            MeilisearchApiError ex = await Assert.ThrowsAsync<MeilisearchApiError>(() => this.defaultClient.GetIndexAsync(indexUid));
-            Assert.Equal("index_not_found", ex.Code);
+            MeilisearchApiError ex = await Assert.ThrowsAsync<MeilisearchApiError>(() => this.client.GetIndexAsync(indexUid));
+            Assert.Equal(ex.Code, "index_not_found");
```
brunoocasali: To solve the GH Actions:

```suggestion
            Assert.Equal("index_not_found", ex.Code);
```

src/Meilisearch/Index.cs
```diff
@@ -404,24 +387,24 @@ public async Task<Settings> GetSettingsAsync(CancellationToken cancellationToken
...
         /// <summary>
         /// Resets all the settings to their default values.
         /// </summary>
         /// <param name="cancellationToken">The cancellation token for this call.</param>
-        /// <returns>Returns the updateID of the asynchronous task.</returns>
-        public async Task<UpdateStatus> ResetSettingsAsync(CancellationToken cancellationToken = default)
+        /// <returns>Returns the task UID of the asynchronous task.</returns>
+        public async Task<TaskInfo> ResetSettingsAsync(CancellationToken cancellationToken = default)
         {
             var httpresponse = await this.http.DeleteAsync($"/indexes/{this.Uid}/settings", cancellationToken).ConfigureAwait(false);
```
brunoocasali:
```suggestion
            var httpResponse = await this.http.DeleteAsync($"/indexes/{this.Uid}/settings", cancellationToken).ConfigureAwait(false);
```

src/Meilisearch/Index.cs
```diff
@@ -404,24 +387,24 @@ public async Task<Settings> GetSettingsAsync(CancellationToken cancellationToken
...
-            return await httpresponse.Content.ReadFromJsonAsync<UpdateStatus>(cancellationToken: cancellationToken).ConfigureAwait(false);
+            return await httpresponse.Content.ReadFromJsonAsync<TaskInfo>(cancellationToken: cancellationToken).ConfigureAwait(false);
```
brunoocasali:
```suggestion
            return await httpResponse.Content.ReadFromJsonAsync<TaskInfo>(cancellationToken: cancellationToken).ConfigureAwait(false);
```

brunoocasali (commented):
Hey, @curquiza I let some ideas for the code, please tell me what you think about them!

Good job! You rocks!

README.md
curquiza: I did not know this convention, and it makes sense!
Could you open an issue so that we can fix it in the future? Definitely a good first issue for the community 😄

src/Meilisearch/TaskEndpoint.cs
curquiza: I'm not sure to understand how to make it `const` since the value changes in the loop

src/Meilisearch/TaskInfo.cs
curquiza: To be honest, I tried to remove what I considered as the useless `set` in these classes, but it ended up with most of the tests failing, so I put back the `set` 😂 I know this is not the good way of coding, but I did not have the time to investigate 😢 But I would love knowing why and how to remove them 
Can we open an issue?

src/Meilisearch/MeilisearchClient.cs
curquiza: I now have this error during the build

Capture d’écran 2022-01-18 à 12 39 34

juchom: this is just a style warning. You can ignore it for now. I suggest that we'll fix this kind of warnings in another pull request where we will patch `.editorconfig` to match the style we want.

There are a few more styles we can change and discuss.

src/Meilisearch/TaskEndpoint.cs
juchom: You want to make `enqueued` and `processing` constants ?

src/Meilisearch/TaskInfo.cs
juchom: We could improve `TaskInfo` to map all the response from the server in another PR.

juchom: But if you want to fix this in this PR just replace `set;` by `init;`

src/Meilisearch/TaskEndpoint.cs
curquiza: Oh ok, my bad I misunderstood, I thought ou wanted to make `task` constant

juchom:
I suggested `TaskInfo` to avoid confusion with `TaskStatus` from the framework, but I'm not sure `TaskInfo` is a good choice.

Leave all class that are prefixed with `Task` to the framework and all the one for this client prefixed with `MTask`

I would probably go with something like `MTaskStatus` or `MeiliTaskStatus` to make things very clear for newcomers...

src/Meilisearch/TaskEndpoint.cs
brunoocasali: Yeah, @juchom what do you think about it?

README.md
brunoocasali: Created https://github.com/meilisearch/meilisearch-dotnet/issues/216

src/Meilisearch/TaskEndpoint.cs
juchom: You could go with something like `ClaimTypes` in framework

https://github.com/dotnet/runtime/blob/main/src/libraries/System.Security.Claims/src/System/Security/Claims/ClaimTypes.cs

curquiza:
Hello @juchom thanks for your involvement 😄 
However, we discussed internally with the team, and we would rather keep the `TaskInfo` name for the moment.
Of course, if we find out you were right, and the name could be clearer, we will update it. Nothing is written in the stone, but we want to let it as it is now for the first version :)

Thanks again! It's definitely interesting to have your opinion here

src/Meilisearch/TaskInfo.cs
curquiza: Thanks @juchom!! :)
Should be fixed in another PR since it's related to all the classes

src/Meilisearch/TaskEndpoint.cs
curquiza: Thanks for the clue! Would be better to do it in another PR :)

src/Meilisearch/MeilisearchClient.cs
curquiza: Would love to have your opinion on it in an issue @juchom! 😄

curquiza:
Checked with @brunoocasali, we merge this PR as it is now. All the possible improvements will be done in another PR pointing to `main` once this package will be compatible with MeiliSearch v0.25.0 (which is the priority now)

PR #217, opened 2022-01-20
Add methods regarding API keys

curquiza:
Add the following methods
- `client.CreateKeyAsync()`
- `client.GetKeyAsync()`
- `client.GetKeysAsync()`
- `client.DeleteKeyAsync()`

⚠️ I did not add `client.UpdateKeysAsync()` because `Http.PatchAsync()` method does not exist for our .Net version

### DIFF
--- a/tests/Meilisearch.Tests/KeyTests.cs
+++ b/tests/Meilisearch.Tests/KeyTests.cs
@@ -29,6 +29,7 @@ namespace Meilisearch.Tests
         {
             var keyResponse = await this.client.GetKeysAsync();
             var keys = keyResponse.Results;
+
             keys.Count().Should().BeGreaterOrEqualTo(2);
         }
 
--- a/tests/Meilisearch.Tests/KeyTests.cs
+++ b/tests/Meilisearch.Tests/KeyTests.cs
@@ -9,8 +9,6 @@ namespace Meilisearch.Tests
     public class KeyTests : IAsyncLifetime
     {
         private IndexFixture fixture;
-        private Key searchKey;
-        private Key adminKey;
         private MeilisearchClient client;
 
         public KeyTests(IndexFixture fixture)
--- a/tests/Meilisearch.Tests/KeyTests.cs
+++ b/tests/Meilisearch.Tests/KeyTests.cs
@@ -1,0 +1,121 @@
+namespace Meilisearch.Tests
+{
+    using System.Linq;
+    using System.Threading.Tasks;
+    using FluentAssertions;
+    using Xunit;
+
+    [Collection("Sequential")]
+    public class KeyTests : IAsyncLifetime
+    {
+        private IndexFixture fixture;
+        private Key searchKey;
+        private Key adminKey;
+        private MeilisearchClient client;
+
+        public KeyTests(IndexFixture fixture)
+        {
+            this.fixture = fixture;
+            this.client = fixture.DefaultClient;
+        }
+
+        public async Task InitializeAsync()
+        {
+            await this.fixture.DeleteAllIndexes(); // Test context cleaned for each [Fact]
+        }
+
+        public Task DisposeAsync() => Task.CompletedTask;
+
+        [Fact]
+        public async Task GetAllKeys()
+        {
+            var keyResponse = await this.client.GetKeysAsync();
+            var keys = keyResponse.Results;
+            keys.Count().Should().BeGreaterOrEqualTo(2);
+        }
+
+        [Fact]
+        public async Task GetOneKeyUsingKeyUid()
+        {
+            var keyResponse = await this.client.GetKeysAsync();
+            var keys = keyResponse.Results;
+            var firstKey = keys.First();
+
+            var fetchedKey = await this.client.GetKeyAsync(firstKey.KeyUid);
+
+            fetchedKey.KeyUid.Should().Equals(firstKey.KeyUid);
+            fetchedKey.Description.Should().Equals(firstKey.Description);
+            fetchedKey.Indexes.Should().Equals(firstKey.Indexes);
+            fetchedKey.Actions.Should().Equals(firstKey.Actions);
+            fetchedKey.ExpiresAt.Should().Equals(firstKey.ExpiresAt);
+            fetchedKey.CreatedAt.Should().Equals(firstKey.CreatedAt);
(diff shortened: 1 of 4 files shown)


tests/Meilisearch.Tests/KeyTests.cs
```diff
@@ -0,0 +1,121 @@
...
+    using System.Threading.Tasks;
+    using FluentAssertions;
+    using Xunit;
+
+    [Collection("Sequential")]
+    public class KeyTests : IAsyncLifetime
+    {
+        private IndexFixture fixture;
+        private Key searchKey;
+        private Key adminKey;
```
curquiza:
```suggestion
```

alallema (approved):
LGTM! 💪 
I guess the `updateKey` method will be added later? It also seems that the linter is not ok with some parts of your code.

tests/Meilisearch.Tests/KeyTests.cs
```diff
@@ -0,0 +1,119 @@
...
+        [Fact]
+        public async Task GetOneKeyUsingKeyUid()
+        {
+            var keyResponse = await this.client.GetKeysAsync();
+            var keys = keyResponse.Results;
+            var firstKey = keys.First();
+
+            var fetchedKey = await this.client.GetKeyAsync(firstKey.KeyUid);
+
+            fetchedKey.KeyUid.Should().Equals(firstKey.KeyUid);
```
brunoocasali, line 44: Asking a dumb question, if we just compare the two instances they will not be equal right? (probably we will need to implement the `IEquatable`) either way, maybe is not the case to just check if the id is the same?

When I started to read multiple implementations of the same code I kind of see some potential for simplification, eg. In this case to ensure that the result is what we expect we just need to check if the response has the type we expect (`Key`) and if the `KeyUid` is what we expect, right? Otherwise, we are testing the implementation of the API (which is probably covered by tests also by the MeiliSearch core team).
I'll learn more about that (and get some references about it) and how I could express myself around that idea, but I just wanted to get some basic answers from you in advance :)

brunoocasali, line 45: Asking a dumb question, if we just compare the two instances they will not be equal right? (probably we will need to implement the `IEquatable`) either way, maybe is not the case to just check if the id is the same?

When I started to read multiple implementations of the same code I kind of see some potential for simplification, eg. In this case to ensure that the result is what we expect we just need to check if the response has the type we expect (`Key`) and if the `KeyUid` is what we expect, right? Otherwise, we are testing the implementation of the API (which is probably covered by tests also by the MeiliSearch core team).
I'll learn more about that (and get some references about it) and how I could express myself around that idea, but I just wanted to get some basic answers from you in advance :)

tests/Meilisearch.Tests/KeyTests.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        public KeyTests(IndexFixture fixture)
+        {
+            this.fixture = fixture;
+            this.client = fixture.DefaultClient;
+        }
+
+        public async Task InitializeAsync()
+        {
+            await this.fixture.DeleteAllIndexes(); // Test context cleaned for each [Fact]
```
brunoocasali, line 22: Shouldn’t we destroy all keys after that also?

tests/Meilisearch.Tests/KeyTests.cs
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        public Task DisposeAsync() => Task.CompletedTask;
+
+        [Fact]
+        public async Task GetAllKeys()
+        {
+            var keyResponse = await this.client.GetKeysAsync();
+            var keys = keyResponse.Results;
+            keys.Count().Should().BeGreaterOrEqualTo(2);
```
brunoocasali, line 33: (just adding a line before the return)
```suggestion

            keys.Count().Should().BeGreaterOrEqualTo(2);
```

brunoocasali (approved):
LGTM! 

I made some comments, but I think nothing should be done right now about it, I think we could discuss internally it  - just wanted to raise the questions around the topic - :)

tests/Meilisearch.Tests/KeyTests.cs
curquiza: Hum, it would be better I think. An issue to think about for every SDK however. 
But this is not an emergency since the keys only impact the tests regarding the keys. But it can definitely bring some corner cases in the future, so I would say, yes, in the future, we should delete the keys we've created after the key tests.

tests/Meilisearch.Tests/KeyTests.cs
curquiza: > eg. In this case to ensure that the result is what we expect we just need to check if the response has the type we expect (Key) and if the KeyUid is what we expect, right? Otherwise, we are testing the implementation of the API (which is probably covered by tests also by the MeiliSearch core team).

Hum, that's not totally true. Some parts have to be tested by the core team only, other parts by the integration team only, and some of them, by "both", even if both teams are not tracking for the same bug.

True story we had in the past:
By mistake, in the ruby SDK, when implementing the `update_stop_words` method, I called the route `POST /indexes/:uid/settings/synonyms` inside the method
Fortunately, I saw it thanks to the CI tests, because I tested two parts
- the test checked the return of the method was correct, here the task information -> impossible to track the mistake here since both actions (adding synonyms and adding stop-words) return the same response.
- but the test also checked, by getting the stop-words, the stop-words were well added -> they were not, so the test failed.

The second test part can seem like a test we should let to the core team only, but it's not -> because you cannot ensure you implemented in the method, you have to ensure the action was processed as expected.
However, following the stop-words example, the integration team should NOT test the search works using the stop words you just added.

Saying this, I agree, this test is maybe not the most relevant for the integration team. Keys are not asynchronous so it's different, but I wanted to take time to explain this anyway :)

I'm happy you start thinking about our way of handling tests, we discussed a lot about it with the team in the past, what is only for the core team or not, using mocking or not... etc 🙂
I'm sure there are many ways to improve our tests to avoid checking the behavior, for example, checking the right HTTP route has been called (`POST /indexes/:uid/settings/stop-words`) rather than checking the required action has been applied (checking the stop-words have been added).

brunoocasali: Thank you for your detailed answer, I'll use that in other opportunities ;)

PR #228, opened 2022-01-27
Remove underscore from private field

brunoocasali:
This should fix this problem: 

`warning SA1309: Field '_taskEndpoint' should not begin with an underscore`


bidoubiwa (approved):
🔥🔥 these were getting so annoying. LGTM

brunoocasali:
bors merge

PR #229, opened 2022-01-27
Enrich nuget data of Meilisearch .net package

brunoocasali:
As the title says, with this data we would have a better description inside the https://www.nuget.org/packages/meilisearch

### DIFF
--- a/src/Meilisearch/Meilisearch.csproj
+++ b/src/Meilisearch/Meilisearch.csproj
@@ -8,6 +8,10 @@
         <Description>.NET wrapper for Meilisearch, an open-source search engine</Description>
         <RepositoryUrl>https://github.com/meilisearch/meilisearch-dotnet</RepositoryUrl>
         <PackageTags>meilisearch;dotnet;sdk;search-engine;search;instant-search</PackageTags>
+        <Authors>Meilisearch</Authors>
+        <PackageProjectUrl>https://meilisearch.com</PackageProjectUrl>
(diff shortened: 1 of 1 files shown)


src/Meilisearch/Meilisearch.csproj
```diff
@@ -8,6 +8,11 @@
         <Description>.NET wrapper for Meilisearch, an open-source search engine</Description>
         <RepositoryUrl>https://github.com/meilisearch/meilisearch-dotnet</RepositoryUrl>
         <PackageTags>meilisearch;dotnet;sdk;search-engine;search;instant-search</PackageTags>
+        <Authors>Meilisearch</Authors>
+        <PackageProjectUrl>https://meilisearch.com</PackageProjectUrl>
+        <PackageIconUrl>https://raw.githubusercontent.com/meilisearch/integration-guides/main/assets/logos/meilisearch_dotnet.png</PackageIconUrl>
+        <PackageIcon>meilisearch_dotnet.png</PackageIcon>
```
bidoubiwa: Do you need to provide the url and the name of the file fetched from the URL?

brunoocasali: Actually, I just copied that from other popular packages (octokit for example), but I think it will raise another problem according to that: https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packing-an-icon-image-file

bidoubiwa (approved):
LGTM :fire:

brunoocasali:
bors merge

PR #233, opened 2022-02-03
Feature/Add a way to use package version programmatically

brunoocasali:
- Create the `Version` class
- Load the current version from the csproj data through the `GetType>Assembly>GetName>Version` (I don’t really know if this could have some potential of not working in some cases, let me know that).

> The Github Action for some reason stopped working, I fixed the version of the .net in order to fix the problem.

Related to https://github.com/meilisearch/integration-guides/issues/150

### DIFF
--- a/src/Meilisearch/Version.cs
+++ b/src/Meilisearch/Version.cs
@@ -11,7 +11,7 @@ namespace Meilisearch
         /// <returns>Returns a formatted version.</returns>
         public string GetQualifiedVersion()
         {
-            return $"Meilisearch (v{this.GetVersion()})";
+            return $"Meilisearch .NET (v{this.GetVersion()})";
         }
 
         /// <summary>
--- a/src/Meilisearch/Version.cs
+++ b/src/Meilisearch/Version.cs
@@ -1,0 +1,26 @@
+namespace Meilisearch
+{
+    /// <summary>
+    /// Information regarding an API key for the Meilisearch server.
+    /// </summary>
+    public class Version
+    {
+        /// <summary>
+        /// Extracts version from Meilisearch.csproj.
+        /// </summary>
+        /// <returns>Returns a formatted version.</returns>
+        public string GetQualifiedVersion()
+        {
+            return $"Meilisearch (v{this.GetVersion()})";
+        }
+
+        /// <summary>
+        /// Extracts the "major.minor.build" version from Meilisearch.csproj.
+        /// </summary>
+        /// <returns>Returns a version from the GetType as String.</returns>
+        public string GetVersion()
+        {
+            return this.GetType().Assembly.GetName().Version.ToString(3);
+        }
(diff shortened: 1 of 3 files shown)


brunoocasali:
bors try

src/Meilisearch/Version.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// </summary>
+    public class Version
+    {
+        /// <summary>
+        /// Extracts version from Meilisearch.csproj.
+        /// </summary>
+        /// <returns>Returns a formatted version.</returns>
+        public string GetQualifiedVersion()
+        {
+            return $"Meilisearch (v{this.GetVersion()})";
```
bidoubiwa: Is this the package version or the version of Meilisearch?

src/Meilisearch/Version.cs
```diff
@@ -0,0 +1,26 @@
...
+        }
+
+        /// <summary>
+        /// Extracts the "major.minor.build" version from Meilisearch.csproj.
+        /// </summary>
+        /// <returns>Returns a version from the GetType as String.</returns>
+        public string GetVersion()
+        {
+            return this.GetType().Assembly.GetName().Version.ToString(3);
```
bidoubiwa, line 23: Just out of curiosity, what does this line do?

src/Meilisearch/Version.cs
brunoocasali: The idea is to retrieve the package version (in this case the dotnet package version) - currently 0.9.0

src/Meilisearch/Version.cs
brunoocasali: Yeah, with this `GetType` call we will be able to get the current object class and after this, we are able to get by reflection the current runtime. From the docs: 

> Represents an assembly, which is a reusable, versionable, and self-describing building block of a common language runtime application. 

And after the `GetName()` which response with an `AssemblyName` I was able to retrieve the current version of the package present in the `Meilisearch.csproj` file.

ref: https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly?view=net-6.0 and
https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assemblyname?view=net-6.0

src/Meilisearch/Version.cs
bidoubiwa: It is not very clear that it is the version of the dot-net package and not the meilisearch engine. 
```suggestion
            return $"Meilisearch NET. (v{this.GetVersion()})";
```
maybe this? 
If you prefer not, no problem at all :) both are good

brunoocasali: Good catch!

bidoubiwa (approved):
LGTM 🔥🔥🔥

brunoocasali:
bors merge

PR #236, opened 2022-02-08
Make sure AddDocuments and UpdateDocuments have the same behaviour

### LINKED ISSUE #231: Issue with null values between AddDocumentsAsync & UpdateDocumentsAsync
**Description**
With actual code, we have an issue with null values in two different parts of the code.

To understand this issue we will do a simple exemple with the following object.

```csharp
public class Person
{
	public int? Age { get; set; }
}
```

Let's see how things are working now :

```csharp
void Main()
{
	var p = new Person { Name = "John", Age = 12 };
	JsonSerializer.Serialize((new List<Person> { p }).RemoveNullValues()).Dump(); // OUTPUT : [{"name":"John","age":12}]
	p.Age = null;
	JsonSerializer.Serialize((new List<Person> { p }).RemoveNullValues()).Dump(); // OUTPUT : [{"name":"John"}]
}
```

The first problem here is the method `RemoveNullValues`, this method is only called in `UpdateDocumentsAsync` and not in `AddDocumentsAsync`.
The second problem is, if we index the whole Person object, the Age field could never be set back to `null`.

I would strongly suggest to delete the `RemoveNullValues`, this method is useless and you will see it next.

Next problem is the default `JsonSerializerOption`

```csharp
public static readonly JsonSerializerOptions JsonSerializerOptions = new JsonSerializerOptions
{
    DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
    PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
};
```

If we modify a little bit the program above like this to use the default `JsonSerializerOptions` like this :
```csharp
void Main()
{
	var p = new Person { Name = "John", Age = 12 };
	JsonSerializer.Serialize((new List<Person> { p }), JsonSerializerOptions).Dump(); // OUTPUT : [{"name":"John","age":12}]
	p.Age = null;
	JsonSerializer.Serialize((new List<Person> { p }), JsonSerializerOptions).Dump(); // OUTPUT : [{"name":"John"}]
}
```

We solved a part of the problem, we don't have to use a custom method to  remove the `null` values, the JsonSerializer is taking care of it.

But we still have the issue if we want to set to `null` the Age field in our index.

What I would suggest is to update the default `JsonSerializerOptions` to

```csharp
public static readonly JsonSerializerOptions JsonSerializerOptions = new JsonSerializerOptions
{
    DefaultIgnoreCondition = JsonIgnoreCondition.Never,
    PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
};
```

And if we run out program again 

If we modify a little bit the program above like to use the default `JsonSerializerOptions` like this :
```csharp
void Main()
{
	var p = new Person { Name = "John", Age = 12 };
	JsonSerializer.Serialize((new List<Person> { p }), JsonSerializerOptions).Dump(); // OUTPUT : [{"name":"John","age":12}]
	p.Age = null;
	JsonSerializer.Serialize((new List<Person> { p }), JsonSerializerOptions).Dump(); // OUTPUT : [{"name":"John","age":null}]
}
```

We are now sending the `null` to the index and we should have the same behavior between `AddDocumentsAsync` and `UpdateDocumentsAsync`

If you're ok with that I will do the PR

juchom:
# Pull Request

## What does this PR do?
Fixes #231 

I've also added two kind of JsonSerializerOptions in `Constants.cs`
One to make sure we remove null values
One to make sure we keep null values

## PR checklist

### DIFF
--- a/src/Meilisearch/Constants.cs
+++ b/src/Meilisearch/Constants.cs
@@ -1,0 +1,29 @@
+﻿namespace Meilisearch
+{
+    using System.Text.Json;
+    using System.Text.Json.Serialization;
+
+    /// <summary>
+    /// This class adds some defaults to work with Meilisearch client.
+    /// </summary>
+    public static class Constants
+    {
+        /// <summary>
+        /// JsonSerializer options used when serializing objects that needs to remove null values.
+        /// </summary>
+        public static readonly JsonSerializerOptions JsonSerializerOptionsRemoveNulls = new JsonSerializerOptions
+        {
+            DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+        };
+
+        /// <summary>
+        /// JsonSerializer options used when serializing objects that keeps null values.
+        /// </summary>
+        public static readonly JsonSerializerOptions JsonSerializerOptionsWriteNulls = new JsonSerializerOptions
+        {
+            DefaultIgnoreCondition = JsonIgnoreCondition.Never,
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+        };
+    }
+}
--- a/src/Meilisearch/Constants.cs
+++ b/src/Meilisearch/Constants.cs
@@ -1,4 +1,4 @@
-﻿namespace Meilisearch
+namespace Meilisearch
 {
     using System.Text.Json;
     using System.Text.Json.Serialization;
--- a/src/Meilisearch/Extensions/HttpExtensions.cs
+++ b/src/Meilisearch/Extensions/HttpExtensions.cs
@@ -74,13 +74,9 @@ namespace Meilisearch.Extensions
             }
         }
 
-        private static StringContent PrepareJsonPayload<T>(T body, JsonSerializerOptions options = default)
+        private static StringContent PrepareJsonPayload<T>(T body, JsonSerializerOptions options = null)
         {
-            if (options == null)
-            {
-                options = new JsonSerializerOptions();
-                options.PropertyNamingPolicy = JsonNamingPolicy.CamelCase;
-            }
+            options ??= Constants.JsonSerializerOptionsWriteNulls;
 
             var payload = new StringContent(JsonSerializer.Serialize(body, options), Encoding.UTF8, "application/json");
             payload.Headers.ContentType.CharSet = string.Empty;
--- a/src/Meilisearch/Index.cs
+++ b/src/Meilisearch/Index.cs
@@ -15,7 +15,7 @@ namespace Meilisearch
     public class Index
     {
         private HttpClient http;
-        private TaskEndpoint _taskEndpoint;
+        private TaskEndpoint taskEndpoint;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="Index"/> class.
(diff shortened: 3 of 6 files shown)


src/Meilisearch/Index.cs
```diff
@@ -15,7 +15,7 @@ namespace Meilisearch
     public class Index
     {
         private HttpClient http;
-        private TaskEndpoint _taskEndpoint;
+        private TaskEndpoint taskEndpoint;
```
brunoocasali: I solved these issues some PR's before, can you update your main branch with our main?

https://github.com/meilisearch/meilisearch-dotnet/blob/main/src/Meilisearch/Index.cs#L18

tests/Meilisearch.Tests/DocumentTests.cs
```diff
@@ -154,8 +154,8 @@ public async Task BasicDocumentsUpdate()
             var docs = await index.GetDocumentsAsync<Movie>();
             Assert.Equal("1", docs.First().Id);
             Assert.Equal("Ironman", docs.First().Name);
+            Assert.Null(docs.First().Genre);
```
brunoocasali, line 157: C# fields are by default null? In this case:

```c#
var movie = new Movie { Id = "1", Name = "Ironman" };

// is the same as 

var movie = new Movie { Id = "1", Name = "Ironman", Genre = null };
```

right?

src/Meilisearch/Extensions/HttpExtensions.cs
```diff
@@ -74,13 +74,9 @@ public static void AddApiKeyToHeader(this HttpClient client, string apiKey)
...
 
-        private static StringContent PrepareJsonPayload<T>(T body, JsonSerializerOptions options = default)
+        private static StringContent PrepareJsonPayload<T>(T body, JsonSerializerOptions options = null)
         {
-            if (options == null)
-            {
-                options = new JsonSerializerOptions();
-                options.PropertyNamingPolicy = JsonNamingPolicy.CamelCase;
-            }
+            options ??= Constants.JsonSerializerOptionsWriteNulls;
```
brunoocasali, line 79: It is good to know C# have this operator is really useful 🥇

brunoocasali (commented):
I just let a comment about testing :)

tests/Meilisearch.Tests/DocumentTests.cs
juchom: Well it depends of the kind of the field.
Reference type are null
Value type have a default value ie: int by default is 0

src/Meilisearch/Index.cs
juchom: Sorry for this one...

src/Meilisearch/Constants.cs
```diff
@@ -0,0 +1,29 @@
...
+        public static readonly JsonSerializerOptions JsonSerializerOptionsRemoveNulls = new JsonSerializerOptions
+        {
+            DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+        };
+
+        /// <summary>
+        /// JsonSerializer options used when serializing objects that keeps null values.
+        /// </summary>
+        public static readonly JsonSerializerOptions JsonSerializerOptionsWriteNulls = new JsonSerializerOptions
```
brunoocasali, line 23: What do you think to use this one inside of the `CreateKeyAsync` method?

src/Meilisearch/MeilisearchClient.cs:315

brunoocasali (approved):
🔥 🚒

brunoocasali:
bors merge

PR #237, opened 2022-02-09
Remove leading slash from api paths

### LINKED ISSUE #198: APIs are called at the root path
**Description**
When providing the `HttpClient` with a `BaseAddress` that has a path such as https://example.com/api/meilisearch, MeiliSearch .NET will call most APIs at the root path regardless - for example https://example.com/health instead of https://example.com/api/meilisearch/health.

**Expected behavior**
MeiliSearch should respect the `BaseAddress` path instead of forcing calls directly against the root.

**Current behavior**
APIs are mostly called like this: `await this.http.GetAsync("/health")` which causes them to always run against the root.
They should be called like this: `await this.http.GetAsync("health")`

There are also isolated cases where it is already done as expected, causing those calls to behave differently than most of the others: https://github.com/meilisearch/meilisearch-dotnet/blob/3137b00ebc62766a75eca947eb990e6904e09295/src/Meilisearch/Index.cs#L98

juchom:
# Pull Request

## What does this PR do?
Fixes #198 

This one is a simple fix, but very important one !

## PR checklist


brunoocasali (commented):
Hello, again @juchom thanks a lot for this PR it will be really useful!

I suggest you add some tests covering the use case regarding this situation:

```cs
this.DefaultClient = new MeilisearchClient("http://localhost:80/api/", "masterKey");
```

and

```cs
this.DefaultClient = new MeilisearchClient("http://localhost:80/api", "masterKey");
```

I ran with these two combinations and I got some interesting results, eg:

`on main` and `/api/` failed only 17 tests.
`on main` and `/api` failed only 5 tests.
`on juchom/fixpath` and `/api/` failed only 5 tests.
`on juchom/fixpath` and `/api` failed only 202 tests.

(I have failing tests in all of the combinations because some of the tests need better setup to ran with a proxied instance I think)
These were the tests who failed in both situations:

- `MeilisearchClientTests.ExceptionWithBadPath`
- `MeilisearchClientTests.BasicUsageOfCustomClient`
- `MeilisearchClientTests.DeleteIndex`
- `MeilisearchClientTests.GetVersionWithCustomClient`
- `MeilisearchClientTests.ErrorHandlerOfCustomClient`

But don't bother yourself with them, the real stuff in this message is the requirement to have a way to pass all the situations with or without the trailing slash because our users could be unpredictable 😄.

Don't hesitate to ask me for better explanations with you need to! 🎸

juchom:
Ok, I went through the tests, I'll try to fix this after the style PR :)

Nextra:
> some of the tests need better setup to ran with a proxied instance I think

Yeah, and from a glance this is true for many other Meilisearch SDKs. I don't think most of the other implemented languages have this particular pitfall, but having Meili behind a reverse proxy is definitely a primary use case that the libraries have to support. I was considering contributing myself for this issue, but really had no idea how to adapt the test setup. Thanks to @juchom for tackling it.

> have a way to pass all the situations with or without the trailing slash because our users could be unpredictable

They shouldn't be, and not all of the situations are supposed to pass. This is just an inherent behavior in .NET because [System.Uri](https://docs.microsoft.com/en-us/dotnet/api/system.uri?view=net-6.0) adheres very strictly to the path merge strategies defined in the [URI RFC3986](https://www.rfc-editor.org/rfc/rfc3986#section-5.2.3). 

Out of the four different combinations - `BaseAddress` with and without trailing slash and `relativeUri` with and without leading slash - only one is usually "correct". And it works exactly the same everywhere the `Uri` class is involved, across the entire .NET ecosystem. I would suggest not breaking with conventions.

As long as this SDK omits the leading slash, it does everything it needs to. The user should know how to pass Meili an `HttpClient` that does what they want it to.

juchom:
I already spoted a few things for the test infra.

I think we can have something working soon.

@brunoocasali @curquiza at meilisearch what reverse proxy would you like to use with docker for intégration tests ?

brunoocasali:
> @brunoocasali @curquiza at meilisearch what reverse proxy would you like to use with docker for intégration tests?

We have some guidelines about it https://docs.meilisearch.com/learn/cookbooks/running_production.html#step-3-secure-and-finish-your-setup-using-a-reverse-proxy-domain-name-and-https

But I think you could use any server you want to do reverse proxies with Meilisearch. 

I have a really simple reverse-proxy config I've created to test your PR:
docker-compose.yml:
```yaml
version: "3.7"

networks:
  intranet:

services:
  nginx: 
    image: nginx:latest
    container_name: production_nginx
    volumes:
      - ./log.log:/etc/nginx/error.log
      - ./nginx.conf:/etc/nginx/nginx.conf
    ports:
      - 80:80
      - 443:443
    networks: 
      - intranet
    depends_on:
      - meilisearch

  meilisearch:
    image: getmeili/meilisearch:v0.25.2
    tty: true
    stdin_open: true
    container_name: meilisearch
    expose:
      - "7700"
    environment:
      - MEILI_MASTER_KEY=masterKey
    networks: 
      - intranet
```

nginx.conf
```config
events {
}

http {
  server {
    server_name _;
    # server_name your.server.url;
    error_log /etc/nginx/error.log debug;

    location /api/ {
      proxy_pass http://meilisearch:7700/;
    }
    
    client_max_body_size 100M;
  }
}
```

brunoocasali:
Hey @juchom you could move this PR to draft if is still in progress :)

juchom:
@brunoocasali FYI I moved your `GetDefaultUserAgentHeader` from `MeilisearchClientTests` to `VersionTests`.

The test was written against a factory only used in test. I have removed the factory because we don't need it, we can just create a single client by hand. This avoid using a third party package `HttpClientFactoryLite`

```csharp
var httpClient = ClientFactory.Instance.CreateClient<MeilisearchClient>();
var client = new MeilisearchClient(httpClient);
```

We can use the instance from the fixture

```csharp
var httpClient = new HttpClient(new MeilisearchMessageHandler(new HttpClientHandler())) { BaseAddress = new Uri(MeilisearchAddress) };
ClientWithCustomHttpClient = new MeilisearchClient(httpClient, ApiKey);
```

I moved it to `VersionTests` because you wanted to check the `AddDefaultUserAgent` method.

juchom:
Ok, now tests are running successfully for the following uris: `http://localhost:7700`, `http://localhost:7700/`, `http://localhost:8080/api/`

This uris is not working `http://localhost:8080/api` because of nginx config. I'm no nginx expert but fixing this and uncomment the file `ProxiedUriWithoutTrailingSlash.cs` should make the test run for every cases.

If anyone can help for the nginx config, it would be great !

juchom:
Not a problem with nginx, but with the `httpclient`

juchom:
Ok, this is almost done. There is one last issue in the `MeilisearchMessageHandler`

It is written as if all response are coming from Meilisearch server. But behind a proxy this not the case anymore. I'm not sure how to deal with this and have the failing unit tests working.

juchom:
Finally all tests are green !

It works for the following uris: `http://localhost:7700`, `http://localhost:7700/`, `http://localhost:8080/api`, `http://localhost:8080/api/`

Let me know if you need help for the review.

Nextra:
> Not a problem with nginx, but with the `httpclient`

I tried to explain this further up: It is not a problem. You don't need to fix it, and it probably shouldn't even be tested like this.

C# developers have to be aware of the trailing slash, it works like this everywhere URIs are involved in .NET. The library is misbehaving if it silently changes ("fixes") the `HttpClient` for you. 

The same goes for adding default headers forcibly instead of injecting them into the requests, if somebody passes an `HttpClient` into this SDK that they are using elsewhere, either of the two will break seemingly at random.

brunoocasali:
> C# developers have to be aware of the trailing slash, it works like this everywhere URIs are involved in .NET. The library is misbehaving if it silently changes ("fixes") the `HttpClient` for you.

So, to be more idiomatic we need to just throw an exception to the user if it sends a "bad" Meilisearch URL? We cannot just let them run their applications and only realize that there is a problem when they tried to use some random method.

juchom:
Actually the issue comes the Uri class.

Creating an Uri with `http://localhost:7700`  transforms it to `http://localhost:7700/` silently.

Creating an Uri with `http://localhost:<http://localhost:7700>8080/api` has no change.

I silently add the trailing slash if not present.

If you prefer an Exception, let me know.
________________________________

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+    container_name: production_nginx
+    volumes:
+      - ./nginx.conf:/etc/nginx/nginx.conf
+    ports:
+      - 8080:80
+    depends_on:
+      - meilisearch
+
+  meilisearch:
+    image: getmeili/meilisearch:v0.25.2
```
brunoocasali: You should use the latest, since this is going to be the default in the CI

src/Meilisearch/MeilisearchClient.cs
```diff
@@ -42,6 +43,7 @@ public MeilisearchClient(string url, string apiKey = default, HttpClient client
         /// <param name="apiKey">API Key to connect to the Meilisearch server. Best practice is to use HttpClient default header rather than this parameter.</param>
         public MeilisearchClient(HttpClient client, string apiKey = default)
         {
+            client.BaseAddress = client.BaseAddress.OriginalString.Trim().EndsWith("/") ? client.BaseAddress : new Uri($"{client.BaseAddress.OriginalString}/");
```
brunoocasali: What about doing this in an extension like the `AddApiKeyToHeader` and the `AddDefaultUserAgent`?

brunoocasali: Could be potentially reused in the line 31.

tests/Meilisearch.Tests/ClientFactory.cs
```diff
@@ -1,34 +0,0 @@
-using System;
```
brunoocasali, line 1: Good change! 👍

tests/Meilisearch.Tests/VersionTests.cs
```diff
@@ -40,5 +43,16 @@ public void GetSimpleVersionFromCsprojFile()
             Assert.NotNull(value);
             Assert.Equal(versionFromCsproj, value);
         }
+
+        [Fact]
+        public void GetDefaultUserAgentHeader()
+        {
+            var httpClient = new HttpClient();
+            httpClient.AddDefaultUserAgent();
```
brunoocasali, line 51: I'm not sure regarding this change, is there any way for us to test the behavior of the `MeilisearchClient`?

Because if someone removes the `AddDefaultUserAgent` from the constructors the tests will not be broken. I was able to test only one constructor at the time I implemented, but now with this change, we are testing none of the usages (from integration test to unit test).

brunoocasali:
> Actually the issue comes the Uri class.

According to @Nextra this is something expected:

> They shouldn't be, and not all of the situations are supposed to pass. This is just an inherent behavior in .NET because [System.Uri](https://docs.microsoft.com/en-us/dotnet/api/system.uri?view=net-6.0) adheres very strictly to the path merge strategies defined in the [URI RFC3986](https://www.rfc-editor.org/rfc/rfc3986#section-5.2.3).


> If you prefer an Exception, let me know.

I like the approach to just raise an exception to the user, and make it realize the mistake. But this will happen only in the runtime, not in the compile time.

Even you mentioning that every C# should know about that @Nextra this kind of problems could happen for example with junior developers or just someone who wants to try Meilisearch and this type of issue could make bad first impressions of the Meilisearch. 

What we try to deliver here in the integrations team, is to deliver more than just a wrapper of the Meilisearch API, but a tool that helps the developers to deliver what they really want, which is their features to their clients, so to make that possible we need to ensure our tools are frictionless.  

@Nextra if we stick to this option (silently add the slash), the main pain point you see is regarding breaking the conventions you mentioned or there is something else?

src/Meilisearch/MeilisearchClient.cs
juchom: Yes we could refactor this. Let me know if you prefer to throw a `BadArgumentException` with a proper message if the `BasseAddress` does not finish with a slash or if we silently add it if missing.

docker-compose.yml
juchom: In your github actions you are tokenizing this version with an environment variable. Is this variable equals to latest ?

tests/Meilisearch.Tests/VersionTests.cs
juchom: Your test actually looks like this : 

https://github.com/meilisearch/meilisearch-dotnet/blob/be67632c21475c060a44b63dfeb8c5c3915a8fbe/tests/Meilisearch.Tests/MeilisearchClientTests.cs#L151-L160

You are only testing the `HttpClient` like the test above. You could remove line 155 and it will still works. I'm not sure how to properly test this with a `MeilisearchClient` right now. Let me know if you want to find a solution for this now or open an issue for later.

Nextra:
> @Nextra if we stick to this option (silently add the slash), the main pain point you see is regarding breaking the conventions you mentioned or there is something else?

Pretty much. Ideally you should touch the `HttpClient` you're given as little as possible. The convention of accepting `HttpClient` is precisely *because* the user wants to configure something outside of the scope of this library. 

Changing settings silently has the risk of other interactions that might dumbfound developers. Not everyone will hand this library a neatly typed `HttpClient` with the factory pattern where all of the changes you perform are isolated (remember: this is a newer extension that does not exist by default in .NET Framework). 

> Even you mentioning that every C# should know about that @Nextra this kind of problems could happen for example with junior developers or just someone who wants to try Meilisearch and this type of issue could make bad first impressions of the Meilisearch.

I absolutely understand where you guys are coming from. This .NET behavior is surprising for someone coming from other languages, which is probably why #198 happened in the first place. However when you're trying to be clever the danger is that _**you**_ are suddenly the only one surprising everyone else.

I would rather "solve" this issue through documentation, but ultimately it is your call. Maybe I'm being too careful in the other direction 😃

tests/Meilisearch.Tests/VersionTests.cs
brunoocasali: I think it's best to create an issue for later, I think we will change the way we deal with headers btw.

docker-compose.yml
brunoocasali: Yes, but that one is just used in the pre-release tests, which means only in the `bump-meilisearch-v*` branches :)

For the other PR's, we should remain into the latest (non-rc) version.

juchom:
What do you think after creating the HttpClient, we check the `BaseAdress` if it ends with a trailing slash. If not we throw a `BadArgumentException` with a clear text on how to fix the issue. So users will know how to fix it.

Also, updating the documentation to make it clear that it is mandatory to have an uri ending with a slash.

docker-compose.yml
juchom: ok, then be careful because I updated two workflows in your github actions.

brunoocasali:
> What do you think after creating the HttpClient, we check the `BaseAdress` if it ends with a trailing slash. If not we throw a `BadArgumentException` with a clear text on how to fix the issue. So users will know how to fix it.
> 
> Also, updating the documentation to make it clear that it is mandatory to have an uri ending with a slash.

Yes, about this @juchom ... I read the comment by Nextra, and I think we could keep the behavior because:

If someone is going to share the same `HttpClient` instance they will need to pass a `BaseAddress` anyway, so if we change it internally is not really a problem. But after the usage of this package, they will somehow go back to the old configuration of the `BaseAddress` in other words, changing the slash is not a problematic change.

But on the other hand, and this I think could be another issue*, is what you mention @Nextra regarding the addition of the headers directly in the `HttpClient`, this definitely is not a good practice and could mess with the following usages of the `HttpClient` in the users' applications.

In summary, @juchom, keep the slash fix and open an issue to fix the usage of the `DefaultRequestHeaders`.

Thank you @Nextra for raising this concerns with us ⭐

src/Meilisearch/MeilisearchClient.cs
brunoocasali: Based on this, https://github.com/meilisearch/meilisearch-dotnet/pull/237#issuecomment-1040314478 I think you could leave as it is, and I'll open a new issue regarding a refactor in the `DefaultRequestHeaders`.

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+    container_name: production_nginx
+    volumes:
+      - ./nginx.conf:/etc/nginx/nginx.conf
+    ports:
+      - 8080:80
+    depends_on:
+      - meilisearch
+
+  meilisearch:
+    image: getmeili/meilisearch:latest
```
brunoocasali: We could use this https://docs.docker.com/compose/compose-file/compose-file-v3/#variable-substitution

```suggestion
    image: "getmeili/meilisearch:${MEILISEARCH_VERSION}"
```

And create a file `.env` with `MEILISEARCH_VERSION=latest`.

According to the compose docs, the `.env` will be used when they cannot find the MEILISEARCH_VERSION env var defined, so this handles the local environment usage and for the tests.yml script.

And in the pre-release script I think this works: `run: MEILISEARCH_VERSION=${{ env.MEILISEARCH_VERSION }} docker-compose up -d` or just `run: docker-compose up -d`

.github/workflows/pre-release-tests.yml
```diff
@@ -21,7 +21,8 @@ jobs:
     - name: Get the latest Meilisearch RC
       run: echo "MEILISEARCH_VERSION=$(curl https://raw.githubusercontent.com/meilisearch/integration-guides/main/scripts/get-latest-meilisearch-rc.sh | bash)" >> $GITHUB_ENV
     - name: Meilisearch (${{ env.MEILISEARCH_VERSION }}) setup with Docker
-      run: docker run -d -p 7700:7700 getmeili/meilisearch:${{ env.MEILISEARCH_VERSION }} ./meilisearch --master-key=masterKey --no-analytics=true
+    #   run: docker run -d -p 7700:7700 getmeili/meilisearch:${{ env.MEILISEARCH_VERSION }} ./meilisearch --master-key=masterKey --no-analytics=true
```
brunoocasali: See this: https://github.com/meilisearch/meilisearch-dotnet/pull/237/files#r806917686

src/Meilisearch/MeilisearchClient.cs
juchom: You can check this file for the extension method `src/Meilisearch/Extensions/StringExtensions.cs`

docker-compose.yml
juchom: It's fixed

brunoocasali:
@juchom I can review everything now? If yes, could you please it make "ready for review" and remove the [WIP] :)

juchom:
I didn't expect to do so much things in this one. I think we can stop here and do the fixing if needed.

.github/workflows/pre-release-tests.yml
```diff
@@ -21,7 +21,8 @@ jobs:
...
+      run: MEILISEARCH_VERSION=${{ env.MEILISEARCH_VERSION }} docker compose up -d
```
brunoocasali:
```suggestion
      run: MEILISEARCH_VERSION=${{ env.MEILISEARCH_VERSION }} docker compose up -d
```

brunoocasali (commented):
Just one small thing remaining! And we are good to go!

juchom:
One last thing I could do to improve testing speed and workflows speed.

Instead of doing all the tests for this urls : `http://localhost:7700` `http://localhost:7700/` `http://localhost:8080/api` `http://localhost:8080/api/`

We can do the tests only for : `http://localhost:7700/` and `http://localhost:8080/api/` and add a few tests for the `ToSafeUri` because a trailing slash will be added for this `http://localhost:7700` and `http://localhost:8080/api`.

And if we go that way, we could spin two Meilisearch containers on 7700 and 7701 (proxied one) and run the tests in parallel.

This is a very easy fix to do.

.github/workflows/pre-release-tests.yml
juchom: Fixed

brunoocasali:
> And if we go that way, we could spin two Meilisearch containers on 7700 and 7701 (proxied one) and run the tests in parallel.
> 
> This is a very easy fix to do.

Wow, awesome @juchom I'll let you implement that so! Let me know when I could review :)

btw: This is the last PR before the next release of this SDK 🥇

juchom:
I added the latest adjustments, let me know if something is not clear.

docker-compose.yml
```diff
@@ -0,0 +1,30 @@
+version: "3.7"
+
+services:
+  meilisearch:
+    image: getmeili/meilisearch:${MEILISEARCH_VERSION}
+    container_name: meilisearch
+    ports:
+      - "7700:7700"
```
brunoocasali, line 8: Have you tried this: 
```yaml
ports:
  - "7701:7700"
  - "7700:7700"
```

Because I think we can have only one service answering in different ports... Or you did this to make possible to work in parallel (to avoid index names conflicts?)

tests/Meilisearch.Tests/StringExtensionsTests.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace Meilisearch.Tests
+{
+    public class StringExtensionsTests
+    {
+        [Theory]
+        [InlineData("http://localhost:7700", "http://localhost:7700/")]
+        [InlineData("http://localhost:7700/", "http://localhost:7700/")]
+        [InlineData("http://localhost:7700/api", "http://localhost:7700/api/")]
+        [InlineData("http://localhost:7700/api/", "http://localhost:7700/api/")]
+        public void CheckUrisEndWithSlash(string actual, string expected)
```
brunoocasali, line 16: I didn't know about this `InlineData` stuff, really useful 👏

tests/Meilisearch.Tests/StringExtensionsTests.cs
```diff
@@ -0,0 +1,19 @@
+using Meilisearch.Extensions;
+
+using Xunit;
+
+namespace Meilisearch.Tests
+{
+    public class StringExtensionsTests
```
brunoocasali, line 9: Could you add a use case test for this condition?

```c#
if (string.IsNullOrWhiteSpace(uri))
{
    throw new ArgumentNullException(nameof(uri));
}
```

docker-compose.yml
juchom: After each tests, data are cleared. So we have to run two instances to avoid conflicts.

We could improve this in the future, to improve tests speed and resource consumptions.

tests/Meilisearch.Tests/StringExtensionsTests.cs
juchom: you have the missing tests

brunoocasali (approved):
Thanks a lot, @juchom for your smart work! 🎉 
Thanks to @Nextra for the help and the guidance! 🎉 

Thanks for contributing with Meilisearch people ❤️ we really appreciate it!

brunoocasali:
bors merge

PR #238, opened 2022-02-09
Project code style

### LINKED ISSUE #216: Adopt a new style guide
Add a new and consistent styling to the project.

- [ ] Adopt new and the most spread style guidelines for C#
- [ ] The new style guide should be guarded with an automated tool that could be integrated with our CI
- [ ] We should have a way to explain to users these new guidelines without bikeshedding.

--- 
## initial issue (before the discussions)
Old issue title: **Make the types explicit instead of using var's everywhere**

This idea came from this code review https://github.com/meilisearch/meilisearch-dotnet/pull/215#discussion_r786077142

According to [Microsoft's good practice guides](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions#implicitly-typed-local-variables) is not good to have `var variable = SomethingYouDontReallyKnowTheReturn();`. 

In order to prevent this from happen this issue talks about removing all of the `var` from:

- [ ] Documentation (eg. code-samples)
- [ ] Source Code
- [ ] Comments
- [ ] Tests

With the concept in mind about **only changing when the type on the right side of the statement is not clear** like the guide says.

Use common sense in order to achieve the best readability possible during the refactor! ;)

juchom:
# Pull Request

## What does this PR do?
Fixes #216 
Revert to .NET Standard 2.0
Set C# language to 7.3
Remove StyleCop

First two rules are :
`using` goes outside namespace
`var` everywhere

The severity is warning right now, this gives a bunch warning during build, we may lower this later.


juchom:
Moving using outside namespace :

FROM:
https://github.com/meilisearch/meilisearch-dotnet/blob/5b758584b6ef62b15c06817670dadf340c0a7f34/src/Meilisearch/Stats.cs#L1-L26


TO:
https://github.com/meilisearch/meilisearch-dotnet/blob/45d90d50d517b79f71b5ee68769cceeede6e2292/src/Meilisearch/Stats.cs#L1-L26

Let me know if you are ok with this style and I will update all the files.

juchom:
You can have a look at `MeilisearchClient.cs`
The `using` are at the top and ordered with `System` first
And we have `var` all the way

Let me know what you think for this two styles before I continue with other options.

src/Meilisearch/MeilisearchClient.cs
```diff
@@ -305,11 +304,11 @@ public async Task<Key> GetKeyAsync(string keyUid, CancellationToken cancellation
         /// <returns>Returns the created API key.</returns>
         public async Task<Key> CreateKeyAsync(Key keyOptions, CancellationToken cancellationToken = default)
         {
-            JsonSerializerOptions jsonOptions = new JsonSerializerOptions
+            var jsonOptions = new JsonSerializerOptions
```
brunoocasali: I think we could use the `Constants.JsonSerializerOptionsWriteNulls` here right?

src/Meilisearch/MeilisearchClient.cs
```diff
@@ -1,17 +1,16 @@
+using System;
```
brunoocasali, line 1: Moving these directives to the outside of the namespace could cause a breaking change for someone?

.editorconfig
```diff
@@ -14,6 +14,19 @@ indent_style             = space
...
+# Sort using and Import directives with System.* appearing first
+dotnet_sort_system_directives_first = true
+
+# 'using' directive preferences
+csharp_using_directive_placement = outside_namespace:warning
+
+# Require var all the time.
+csharp_style_var_for_built_in_types = true:warning
+csharp_style_var_when_type_is_apparent = true:warning
+csharp_style_var_elsewhere = true:warning
```
brunoocasali: For your current changes in this PR, making this `csharp_style_var_elsewhere = false:warning`, would change something?

src/Meilisearch/MeilisearchClient.cs
juchom: No, this is fine for this library.

If you really want to understand why this option exists : https://stackoverflow.com/a/151560/1026105

The default nowadays is to have them outside.

.editorconfig
juchom: This would look like this

```csharp
// csharp_style_var_elsewhere = true
var f = this.Init();

// csharp_style_var_elsewhere = false
bool f = this.Init();
```

src/Meilisearch/MeilisearchClient.cs
juchom: Fixed

juchom:
Ok, well actually it will probably easier if you look at the project with all changes.

It will take too much time to discuss every rules.

This is more or less how an OSS project looks like nowadays. Let me know if you have any questions.

src/Meilisearch/MeilisearchClient.cs
```diff
@@ -53,7 +53,7 @@ public MeilisearchClient(HttpClient client, string apiKey = default)
         /// <returns>Returns the Meilisearch version with commit and build version.</returns>
         public async Task<MeiliSearchVersion> GetVersionAsync(CancellationToken cancellationToken = default)
         {
-            var response = await this.http.GetAsync("/version", cancellationToken).ConfigureAwait(false);
+            var response = await _http.GetAsync("/version", cancellationToken).ConfigureAwait(false);
```
brunoocasali, line 56: So now `this.` is the default way to lookup for the references C# does?

brunoocasali:
@juchom sorry, I clicked on the ready for review button, sorry about this!

brunoocasali (changes requested):
I saw your entire PR and really liked the changes, even if I'm more explicit regarding the var/type subject.

The whole point over the generated issue is to have a standard to develop this package and prevent bikeshedding discussions!

But I still have questions regarding some subjects:

- Should we use warnings or errors? (I prefer errors, but this is possible or we will need to refactor the entire src?)
- Since there is no more `StyleCop`, do we still have the power to use some kind of lint-checker into our CI?

src/Meilisearch/MeilisearchClient.cs
juchom: Most C# OSS library and asp.net core code looks like this.

juchom:
> * Should we use warnings or errors? (I prefer errors, but this is possible or we will need to refactor the entire src?)

You can chose whatever you want and adjust it in the futur. Some OSS authors allows to merge in `main` with warnings, but treats warning as errors when they publish a new package.

This can be adjusted with `.editorconfig` in the futur. You define your workflow and what is an error, a warning, a suggestion or what should be silenced.

> * Since there is no more StyleCop, do we still have the power to use some kind of lint-checker into our CI?

I removed `StyleCop` because a good part of it can be managed with `.editorconfig` directly. `StyleCop` was relevant before `.editorconfig` was available, nowadays we don't see much project using it.

So `.editorconfig` with `dotnet format` is already doing this. If you want more control we can add some rolsyn analyzers, `StyleCop` is one, but there are many of them. I would not introduce them now and wait to see if you have some recuring pain points that can be managed with analyzers.

brunoocasali:
@juchom do you want to change something else? I'm ok to accept!

juchom:
You can merge it. The main idea is to have a code that looks like what we can find.

If you miss something or want to change something we can change it later, it won't be that big.

EDIT: You're not mixing `rulesets` and `editorconfig` anymore.

brunoocasali (approved):
LGTM! Thanks for your time @juchom 🎉

brunoocasali:
bors merge

PR #245, opened 2022-02-16
Add UpdateKey method

### LINKED ISSUE #218: Implement `client.UpdateKeyAsync()` method
With the release of [MeiliSearch v0.25.0](https://github.com/meilisearch/MeiliSearch/releases/tag/v0.25.2), and following this PR (https://github.com/meilisearch/meilisearch-dotnet/pull/217), I was not able to implement `client.UpdateKeyAsync()` method.

Looks like `Http.PatchAsync` is not available with the .NET version we are using.

I might have missed something 🙂

juchom:
# Pull Request

## What does this PR do?
Fixes #218 

## PR checklist

### DIFF
--- a/src/Meilisearch/ApiActions.cs
+++ b/src/Meilisearch/ApiActions.cs
@@ -6,7 +6,7 @@ namespace Meilisearch
     public static class ApiActions
     {
         /// <summary>
-        /// Authorize all actions
+        /// Authorize all actions.
         /// </summary>
         public const string All = "*";
 
--- a/src/Meilisearch/ApiActions.cs
+++ b/src/Meilisearch/ApiActions.cs
@@ -1,0 +1,85 @@
+namespace Meilisearch
+{
+    /// <summary>
+    /// List all available actions for a key.
+    /// </summary>
+    public static class ApiActions
+    {
+        /// <summary>
+        /// Authorize all actions
+        /// </summary>
+        public const string All = "*";
+
+        /// <summary>
+        /// Provides access to both POST and GET search endpoints on authorized indexes.
+        /// </summary>
+        public const string Search = "search";
+
+        /// <summary>
+        /// Provides access to the add documents and update documents endpoints on authorized indexes.
+        /// </summary>
+        public const string DocumentsAdd = "documents.add";
+        /// <summary>
+        /// Provides access to the get one document and get documents endpoints on authorized indexes.
+        /// </summary>
+        public const string DocumentsGet = "documents.get";
+        /// <summary>
+        /// Provides access to the delete one document, delete all documents, and batch delete endpoints on authorized indexes.
+        /// </summary>
+        public const string DocumentsDelete = "documents.delete";
+
+        /// <summary>
+        /// Provides access to the create index endpoint.
+        /// </summary>
+        public const string IndexesCreate = "indexes.create";
+        /// <summary>
+        /// Provides access to the get one index and list all indexes endpoints.
+        /// Non-authorized indexes will be omitted from the response.
+        /// </summary>
+        public const string IndexesGet = "indexes.get";
+        /// <summary>
+        /// Provides access to the update index endpoint.
+        /// </summary>
+        public const string IndexesUpdate = "indexes.update";
+        /// <summary>
+        /// Provides access to the delete index endpoint.
+        /// </summary>
+        public const string IndexesDelete = "indexes.delete";
+
+        /// <summary>
+        /// Provides access to the get one task and get all tasks endpoints.
+        /// Tasks from non-authorized indexes will be omitted from the response.
+        /// Also provides access to the get one task by index and get all tasks by index endpoints on authorized indexes.
+        /// </summary>
+        public const string TasksGet = "tasks.get";
+
+        /// <summary>
+        /// Provides access to the get settings endpoint and equivalents for all subroutes on authorized indexes.
+        /// </summary>
+        public const string SettingsGet = "settings.get";
+        /// <summary>
+        /// Provides access to the update settings and reset settings endpoints and equivalents for all subroutes on authorized indexes.
+        /// </summary>
+        public const string SettingsUpdate = "settings.update";
+
+        /// <summary>
+        /// Provides access to the get stats of an index endpoint and the get stats of all indexes endpoint.
+        /// For the latter, non-authorized indexes are omitted from the response.
+        /// </summary>
+        public const string StatsGet = "stats.get";
+
+        /// <summary>
+        /// Provides access to the create dump endpoint. Not restricted by indexes.
+        /// </summary>
+        public const string DumpsCreate = "dumps.create";
+        /// <summary>
+        /// Provides access to the get dump status endpoint. Not restricted by indexes.
+        /// </summary>
+        public const string DumpsGet = "dumps.get";
+
+        /// <summary>
+        /// Provides access to the get Meilisearch version endpoint.
+        /// </summary>
+        public const string Version = "version";
+    }
+}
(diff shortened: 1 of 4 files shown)


juchom:
According to the documentation on updating a key :

> Update the description, permissions, or expiration date of an API key. To learn more about the variables sent in the body of the request, see the [create key](https://docs.meilisearch.com/reference/api/keys.html#body) endpoint.


> Updates to keys are partial. This means you should provide only the fields you intend to update, as any fields not present in the payload will remain unchanged.

We have to be careful with strongly typed language and the expiry date of a key.

The JsonSerializerOptions removes null values which is fine for all fields except the `ExpiresAt`. Setting this field to null has a real meaning (not expiring).

Right now I have modified the Key object like this :

```csharp
public class Key
{
    /// <summary>
    /// Create a key whith an expiry date. Pass null if it's not expiring.
    /// </summary>
    /// <param name="expiresAt">Expiry date.</param>
    public Key(DateTime? expiresAt)
    {
        ExpiresAt = expiresAt;
    }

    /// <summary>
    /// Gets or sets the date when the API key expires.
    /// </summary>
    [JsonIgnore(Condition = JsonIgnoreCondition.Never)]
    public DateTime? ExpiresAt { get; private set; }

    /// <summary>
    /// Update the key expiry date to a new one. Pass null if you want it to never expires.
    /// </summary>
    /// <param name="expiresAt">New expiry date.</param>
    public void SetExpiresAt(DateTime? expiresAt)
    {
        ExpiresAt = expiresAt;
    }
}
```

I added a new constructor where you have to pass the ExpiryDate
I decorated the ExpiresAt field with ` [JsonIgnore(Condition = JsonIgnoreCondition.Never)]` in order to tell the serializer to always send this value even if it's null.
I added the method `public void SetExpiresAt(DateTime? expiresAt)` to make sure that if a user want to change the expiry date he understands what he is going to do.

Do you think the developer experience will be good or do you have a better idea ?

sander1095:
I think it'd be good to seperate the API actions, nullable expires at and updateKey into seperate PR's if possible :) Nice and clean and revertable. Otherwise it looks good to me I think!

juchom:
> I think it'd be good to seperate the API actions, nullable expires at and updateKey into seperate PR's if possible :) Nice and clean and revertable. Otherwise it looks good to me I think!

This is an option, I'm not sure this is the way to go.

@curquiza @brunoocasali according to Meiliseach documentation, if a key has a null value for ExpiresAt, it means that it will never expires.
Why not having an ExpiresAt value far in the future for this instead of a null values ? JsonSerializer can be tricky for other clients with strongly typed languages.
Forcing a value to be defined on the server would make it clear for everyone on when the key will expire. (Azure portal manages keys like that)

src/Meilisearch/ApiActions.cs
```diff
@@ -0,0 +1,85 @@
+namespace Meilisearch
+{
+    /// <summary>
+    /// List all available actions for a key.
+    /// </summary>
+    public static class ApiActions
+    {
+        /// <summary>
+        /// Authorize all actions
```
brunoocasali:
```suggestion
        /// Authorize all actions.
```

src/Meilisearch/ApiActions.cs
```diff
@@ -0,0 +1,85 @@
+namespace Meilisearch
+{
+    /// <summary>
+    /// List all available actions for a key.
+    /// </summary>
+    public static class ApiActions
```
brunoocasali, line 6: Is there a particular reason to not use an enum here?

src/Meilisearch/Key.cs
```diff
@@ -9,6 +9,15 @@ namespace Meilisearch
     /// </summary>
     public class Key
     {
+        /// <summary>
+        /// Create a key whith an expiry date. Pass null if it's not expiring.
```
brunoocasali:
```suggestion
        /// Create a key with an expiry date. Pass null if it's not expiring.
```

src/Meilisearch/Key.cs
```diff
@@ -44,5 +54,14 @@ public class Key
         /// Gets or sets the date when the API key was updated.
         /// </summary>
         public DateTime UpdatedAt { get; set; }
+
+        /// <summary>
+        /// Update the key expiry date to a new one. Pass null if you want it to never expires.
```
brunoocasali, line 59:
```suggestion
        /// Update the key expiry date to a new one. Pass null if you want it to never expire.
```

src/Meilisearch/Key.cs
```diff
@@ -9,6 +9,15 @@ namespace Meilisearch
     /// </summary>
     public class Key
     {
+        /// <summary>
+        /// Create a key whith an expiry date. Pass null if it's not expiring.
+        /// </summary>
+        /// <param name="expiresAt">Expiry date.</param>
+        public Key(DateTime? expiresAt)
```
brunoocasali, line 16: I read your comment regarding this change, and I understood your intentions about that, but I am not sure if this is the way to go. I like the approach to make it harder to user make a mess, but you know they still, can make a mess. 

Given that we released v0.10 a few weeks ago, I think we could wait for more feedback before introducing a new breaking change.

> Why not having an ExpiresAt value far in the future for this instead of a null values ? JsonSerializer can be tricky for other clients with strongly typed languages.

I partially agree because I don't like the idea to have a key that never expires, this always triggers a security concern. But since we have this option, I agree with the idea to have the null value to represent it.

I invite you to open a discussion here: https://github.com/meilisearch/product/discussions regarding this topic, I'll be glad to share my concerns there :)

brunoocasali (commented):
Hello, @juchom thanks a lot for your PR (as always 🥇)!

I left some comments regarding the implementation, I think we are on the right path, but I want to let things cold a little bit before introducing this new breaking change.

src/Meilisearch/MeilisearchClient.cs
```diff
@@ -314,6 +315,22 @@ await _http.PostAsJsonAsync("keys", keyOptions, Constants.JsonSerializerOptionsW
...
+        /// Updates an API key for the Meilisearch server.
+        /// </summary>
+        /// <param name="keyUid">Unique identifier of the API key.</param>
+        /// <param name="keyOptions">The options of the API key.</param>
+        /// <param name="cancellationToken">The cancellation token for this call.</param>
+        /// <returns>Returns the created API key.</returns>
+        public async Task<Key> UpdateKeyAsync(string keyUid, Key keyOptions, CancellationToken cancellationToken = default)
+        {
+            var json = JsonSerializer.Serialize(keyOptions, Constants.JsonSerializerOptionsRemoveNulls);
+            var request = new HttpRequestMessage(new HttpMethod("PATCH"), $"keys/{keyUid}");
```
brunoocasali, line 328: Is this a good suggestion in this case? I like this way to instantiate objects 🏒 

```c#
var request = new HttpRequestMessage(new HttpMethod("PATCH"), $"keys/{keyUid}") {
       Content = new StringContent(json, Encoding.UTF8, "application/json")
};
```

src/Meilisearch/ApiActions.cs
juchom: In C# enum cannot be mapped directly to string. We can use attributes to decorate them and extract the value of it. I think it's a bit hacky to be honest.

src/Meilisearch/MeilisearchClient.cs
juchom: This is fine for me

juchom:
I had a nice discussion with @gmourier about this issue.

I'll close it for now and recreate one when the last arbitrages will be made.

PR #264, opened 2022-05-18
Jc/mini cleanup

juchom:
# Pull Request

## What does this PR do?
This is mostly a cosmetic cleanup.

The `Constants.cs` is now marked as internal. There is no need to surface this to the end-user. We don't want them to use it their own project. We want to be able to change what is in this file without creating breaking changes in the future.

The `Index.cs` was pretty huge, I've splitted it in smaller files that are big sections of the definition of an index and document management. This will allow new comers to study the code easily and make refactoring, upgrades, maintenance easier for developers.

image

Let me know if you like it.


src/Meilisearch/Meilisearch.csproj
```diff
@@ -17,4 +17,28 @@
     <ItemGroup>
       <PackageReference Include="System.Net.Http.Json" Version="6.0.0" />
     </ItemGroup>
+
+    <ItemGroup>
```
brunoocasali, line 22: Just a curiosity, why do you need to make this change as well? The `partial` keyword doesn't work out of the box?

brunoocasali (dismissed):
Thanks a lot for your contribution @juchom you rocks! 👨‍🎤 

Meilisearch is better with contributors like you!

PS: I'll let @alallema review as well since she is the new official maintainer of this package :)

src/Meilisearch/Meilisearch.csproj
juchom: >Just a curiosity, why do you need to make this change as well?

The csproj modification is for the nesting like this

image

Without the csproj modification it would looks like this
image

>The `partial` keyword doesn't work out of the box?

The `partial` keyword is mandatory if you want to split a class across several files.

brunoocasali: Oh nice, good to know!

alallema (commented):
Hi @juchom,
Thanks a lot for this PR is really great I love it. I just have a hesitation about the settings, I wonder if the following files (`Index.Attributes.cs`, `Index.RankingsRules.cs`, `Index.StopWords.cs` and `Index.Synonyms.cs`) should be together in the `Index.Settings.cs` because there are all related to the Settings what do you think?

juchom:
Hi @alallema,

Good question, I had the exact same one when I authored this PR.

The main issue I see with having all the Settings in one file is there is a lot of Settings
DisplayableAttributes
DistinctAttributes
FilterableAttributes
SearchAttribultes
SortableAttributes
All the attributes above (198 lines)

RankingRules (47 lines)

StopWords (47 lines)

Synonyms (48 lines)

While writing this reply, I think we could merge all of this in one file as you suggest. The settings part of the client is not moving that much.

I think it would be best to merge everything in a Index.Settings.cs file. Let me know if you prefer this solution.

alallema:
> The main issue I see with having all the Settings in one file is there are a lot of Settings
> The settings part of the client is not moving that much.

I agree with you, and indeed this part would not change but others will come so maybe it is better to leave as you did.
My bad I could have approved directly 😊

alallema (dismissed):
LGTM! 🔥 🔥 🔥 
Thanks again for this PT @juchom

alallema:
@juchom, did you prefer to handle the conflict or I can to do it myself?

juchom:
> @juchom, did you prefer to handle the conflict or I can to do it myself?

Let me handle this !

alallema (dismissed):
LGTM! Perfect 🎉

alallema (approved):

alallema:
bors merge

PR #269, opened 2022-05-24
Add support for JSON string

juchom:
# Pull Request

## What does this PR do?
Fixes JSON string support in #182

This adds support for JSON string Add and Update.

For the tests I introduced the sample files used in other repo.


alallema (commented):
Thanks, @juchom for this PR ❤️ !!!
I really like the new Dataset directory, but did we need all the files? I don't think the `movie.json` is required also there is already a nested index in the [tests/Meilisearch.Tests/IndexFixture.cs](https://github.com/meilisearch/meilisearch-dotnet/blob/5aa7ec9992c19f329c29ee3b85dd3c143765d0b8/tests/Meilisearch.Tests/IndexFixture.cs#L137) however it could be a good enhancement to use all these files instead of these index declaration but maybe not in this PR. What do you think?

juchom:
> I really like the new Dataset directory, but did we need all the files?

If we are strict, we only need one for this PR and we can remove all others.

The others do not make sense at the moment, but we will use them for sure in new unit tests or updating existing ones.

Having them ready to use in our tests is also a way to let the contributors know this files are available to use in their tests.

alallema:
> Having them ready to use in our tests is also a way to let the contributors know this files are available to use in their tests.

I agree with you, it would be nice if contributors could find these files but I prefer to create an issue about it with the link to the files if it's ok with you. I can create the issue unless you prefer to do so.

juchom:
Ok, I'll let you open the issue for others files and I only keep the `small_movies.json` in this PR.

alallema (approved):
LGTM! 🎉 
Thanks again @juchom for this great work

alallema:
bors merge

PR #275, opened 2022-05-31
Reduce public api surface

juchom:
# Pull Request

## What does this PR do?

At the moment, there are too many functions that are not supposed to be used by your customers.

This make sure people won't use code that is not supposed to be surfaced and avoid breaking changes.


alallema (approved):
Thanks for this PR @juchom it is indeed a great modification 🎉 
LGTM!

alallema:
bors merge

PR #280, opened 2022-06-09
Make SearchResult an immutable object

juchom:
# Pull Request

## What does this PR do?
This PR continues to reduce public API surface and make sure we map the payload sent by Meilisearch.

- Documentation have been updated
- Properties are read only and ordered like in the documentation
- `IEnumerable` are now `IReadOnlyCollection` 
- Added `ProcessingTimeMs` property

Why `IEnumerable` to `IReadOnlyCollection` : 
https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/guidelines-for-collections?redirectedfrom=MSDN

> DO use ReadOnlyCollection<T>, a subclass of ReadOnlyCollection<T>, or in rare cases IEnumerable<T> for properties or return values representing read-only collections.

This will also avoid the warning message about multiple evaluation due to potential deferred execution and avoid to use the `.ToList()`.

### DIFF
--- a/src/Meilisearch/SearchResult.cs
+++ b/src/Meilisearch/SearchResult.cs
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using System.Text.Json.Serialization;
 
 namespace Meilisearch
 {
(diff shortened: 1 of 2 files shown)


alallema (approved):
Thanks, @juchom for this PR !!! 🔥 
LGTM! And sorry for the delay as usual 😄

PR #287, opened 2022-07-12
Add PatchAsync PatchAsJsonAsync methods

### LINKED ISSUE #218: Implement `client.UpdateKeyAsync()` method

alallema:
# Pull Request

## What does this PR do?
Fixes #218

### DIFF
--- a/src/Meilisearch/Extensions/HttpExtensions.cs
+++ b/src/Meilisearch/Extensions/HttpExtensions.cs
@@ -92,10 +92,9 @@ namespace Meilisearch.Extensions
             return payload;
         }
 
-        private static Task<HttpResponseMessage> PatchAsync(this HttpClient client, string? requestUri, HttpContent content, CancellationToken cancellationToken)
+        private static Task<HttpResponseMessage> PatchAsync(this HttpClient client, string requestUri, HttpContent content, CancellationToken cancellationToken)
         {
-            var uri = string.IsNullOrEmpty(requestUri) ? null : new Uri(requestUri, UriKind.RelativeOrAbsolute);
-            return client.PatchAsync(uri, content, cancellationToken);
+            return client.PatchAsync(requestUri, content, cancellationToken);
         }
 
         internal static Task<HttpResponseMessage> PatchAsJsonAsync<TValue>(this HttpClient client, string? requestUri, TValue value, JsonSerializerOptions? options = null, CancellationToken cancellationToken = default)
--- a/src/Meilisearch/Extensions/HttpExtensions.cs
+++ b/src/Meilisearch/Extensions/HttpExtensions.cs
@@ -1,5 +1,7 @@
+using System;
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Net.Http.Json;
 using System.Text;
 using System.Text.Json;
 using System.Threading;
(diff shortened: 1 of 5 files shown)


src/Meilisearch/Extensions/HttpExtensions.cs
```diff
@@ -81,13 +83,49 @@ internal static void AddDefaultUserAgent(this HttpClient client)
...
+        }
+
+        internal static Task<HttpResponseMessage> PatchAsJsonAsync<TValue>(this HttpClient client, string? requestUri, TValue value, JsonSerializerOptions? options = null, CancellationToken cancellationToken = default)
+        {
+            if (client == null)
+            {
+                throw new ArgumentNullException(nameof(client));
+            }
+
+            var content = JsonContent.Create(value, mediaType: null, options);
```
brunoocasali: I think you should use the `PrepareJsonPayload` don't you?

like this https://github.com/meilisearch/meilisearch-dotnet/blob/b8f0fe37174183d93e25d1305ceefa352b05608f/src/Meilisearch/Extensions/HttpExtensions.cs#L28

src/Meilisearch/Extensions/HttpExtensions.cs
```diff
@@ -81,13 +83,49 @@ internal static void AddDefaultUserAgent(this HttpClient client)
...
+            var method = new HttpMethod("PATCH");
+            var request = new HttpRequestMessage(method, requestUri) { Content = content };
+            return client.SendAsync(request, cancellationToken);
+        }
+
+        internal static Task<HttpResponseMessage> PatchAsJsonAsync<TValue>(this HttpClient client, string? requestUri, TValue value, JsonSerializerOptions? options = null, CancellationToken cancellationToken = default)
+        {
+            if (client == null)
+            {
+                throw new ArgumentNullException(nameof(client));
```
brunoocasali: Is that a valid validation, since this will be an extension, a valid instance of MeilisearchClient is required to be able to call this method...

src/Meilisearch/Extensions/HttpExtensions.cs
```diff
@@ -81,13 +83,49 @@ internal static void AddDefaultUserAgent(this HttpClient client)
...
+            }
+
+            var content = JsonContent.Create(value, mediaType: null, options);
+            return client.PatchAsync(requestUri, content, cancellationToken);
+        }
+
+        internal static Task<HttpResponseMessage> PatchAsJsonAsync<TValue>(this HttpClient client, Uri? requestUri, TValue value, JsonSerializerOptions? options = null, CancellationToken cancellationToken = default)
```
brunoocasali: Why do you need an overload for this method? I think one option (string or Uri) will be fine, no?

alallema: Yes, but the [real one](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.patchasync?view=net-6.0) takes both so I couldn't choose

src/Meilisearch/Extensions/HttpExtensions.cs
alallema: I hesitate but this method is used for the custom method we made from  HTTP-`JsonCustomAsync` so I preferred to keep the basic implementation of what the function should be in case one day we will use the real one.
Don't you think?

src/Meilisearch/Extensions/HttpExtensions.cs
brunoocasali: Actually, I think you can opt for just the `string` one, this file is just a collection of methods that will add logic to the instance of a `HttpClient`. In other words this is not a `Interface` where you cannot leave without implementing all the methods that interface requires :) 

So it will be ok to remove the `Uri` one :)

src/Meilisearch/Extensions/HttpExtensions.cs
```diff
@@ -81,13 +83,39 @@ internal static void AddDefaultUserAgent(this HttpClient client)
...
             payload.Headers.ContentType.CharSet = string.Empty;
 
             return payload;
         }
+
+        private static Task<HttpResponseMessage> PatchAsync(this HttpClient client, string? requestUri, HttpContent content, CancellationToken cancellationToken)
+        {
+            var uri = string.IsNullOrEmpty(requestUri) ? null : new Uri(requestUri, UriKind.RelativeOrAbsolute);
+            return client.PatchAsync(uri, content, cancellationToken);
+        }
```
brunoocasali: Couple of things:

- is there any case in which the `requestUri` will not be sent?
- since this is an internal method, it sounds redundant to validate something we know should be sent (I mean, the end-user has no power over the input in this case).
- if you want to keep the validation you should remove the version with `Uri`  (`PatchAsJsonAsync<TValue>(this HttpClient client, Uri? requestUri, TValue `)

```suggestion
        private static Task<HttpResponseMessage> PatchAsync(this HttpClient client, string requestUri, HttpContent content, CancellationToken cancellationToken)
        {
            return client.PatchAsync(requestUri, content, cancellationToken);
        }
```

brunoocasali (approved):
🎉 🎉 perfection! 🤌

PR #288, opened 2022-07-12
Apply tasks changes

alallema:
# Pull Request

Related to: https://github.com/meilisearch/integration-guides/issues/205

## What does this PR do?

### Changes

- Remove `GET /indexes/:indexUid/tasks`. Use `GET /tasks?indexUid=:indexUid` instead.
- Remove `GET /indexes/:indexUid/tasks/:taskUid`. Use `GET /tasks/:taskUid` instead.
- Add new class `TaskRessource` return by:
  - `client.GetTaskAsync`
  - `client.GetTasksAsync`
  - `index.GetTaskAsync`
  - `index.GetTasksAsync`
- Remove `GetIndexTasksAsync` method
- Rename `uid` field into `taskUid` in `TaskInfo` class

### DIFF
--- a/src/Meilisearch/Index.Tasks.cs
+++ b/src/Meilisearch/Index.Tasks.cs
@@ -11,9 +11,9 @@ namespace Meilisearch
         /// </summary>
         /// <param name="cancellationToken">The cancellation token for this call.</param>
         /// <returns>Returns a list of the operations status.</returns>
-        public async Task<Result<IEnumerable<TaskInfo>>> GetTasksAsync(CancellationToken cancellationToken = default)
+        public async Task<TasksResults<IEnumerable<TaskRessource>>> GetTasksAsync(CancellationToken cancellationToken = default)
         {
-            return await TaskEndpoint().GetIndexTasksAsync(Uid, cancellationToken).ConfigureAwait(false);
+            return await TaskEndpoint().GetTasksAsync(cancellationToken).ConfigureAwait(false);
         }
 
         /// <summary>
@@ -22,9 +22,9 @@ namespace Meilisearch
         /// <param name="taskUid">Uid of the task.</param>
         /// <param name="cancellationToken">The cancellation token for this call.</param>
         /// <returns>Returns the tasks.</returns>
-        public async Task<TaskInfo> GetTaskAsync(int taskUid, CancellationToken cancellationToken = default)
+        public async Task<TaskRessource> GetTaskAsync(int taskUid, CancellationToken cancellationToken = default)
         {
-            return await TaskEndpoint().GetIndexTaskAsync(Uid, taskUid, cancellationToken).ConfigureAwait(false);
+            return await TaskEndpoint().GetTaskAsync(taskUid, cancellationToken).ConfigureAwait(false);
         }
 
         /// <summary>
@@ -35,7 +35,7 @@ namespace Meilisearch
         /// <param name="intervalMs">Interval in millisecond between each check.</param>
         /// <param name="cancellationToken">The cancellation token for this call.</param>
         /// <returns>Returns the task info of finished task.</returns>
-        public async Task<TaskInfo> WaitForTaskAsync(
+        public async Task<TaskRessource> WaitForTaskAsync(
             int taskUid,
             double timeoutMs = 5000.0,
             int intervalMs = 50,
--- a/src/Meilisearch/Index.Tasks.cs
+++ b/src/Meilisearch/Index.Tasks.cs
@@ -11,7 +11,7 @@ namespace Meilisearch
         /// </summary>
         /// <param name="cancellationToken">The cancellation token for this call.</param>
         /// <returns>Returns a list of the operations status.</returns>
-        public async Task<TasksResults<IEnumerable<TaskRessource>>> GetTasksAsync(CancellationToken cancellationToken = default)
+        public async Task<TasksResults<IEnumerable<TaskResource>>> GetTasksAsync(CancellationToken cancellationToken = default)
         {
             return await TaskEndpoint().GetTasksAsync(cancellationToken).ConfigureAwait(false);
         }
@@ -22,7 +22,7 @@ namespace Meilisearch
         /// <param name="taskUid">Uid of the task.</param>
         /// <param name="cancellationToken">The cancellation token for this call.</param>
         /// <returns>Returns the tasks.</returns>
-        public async Task<TaskRessource> GetTaskAsync(int taskUid, CancellationToken cancellationToken = default)
+        public async Task<TaskResource> GetTaskAsync(int taskUid, CancellationToken cancellationToken = default)
         {
             return await TaskEndpoint().GetTaskAsync(taskUid, cancellationToken).ConfigureAwait(false);
         }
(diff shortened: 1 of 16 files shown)


src/Meilisearch/Index.Tasks.cs
```diff
@@ -22,9 +22,9 @@ public async Task<Result<IEnumerable<TaskInfo>>> GetTasksAsync(CancellationToken
         /// <param name="taskUid">Uid of the task.</param>
         /// <param name="cancellationToken">The cancellation token for this call.</param>
         /// <returns>Returns the tasks.</returns>
-        public async Task<TaskInfo> GetTaskAsync(int taskUid, CancellationToken cancellationToken = default)
+        public async Task<TaskRessource> GetTaskAsync(int taskUid, CancellationToken cancellationToken = default)
```
brunoocasali: It should be `TaskResource` :)
image

src/Meilisearch/TaskEndpoint.cs
```diff
@@ -43,24 +43,12 @@ public async Task<TaskInfo> GetTaskAsync(int taskUid, CancellationToken cancella
...
         }
 
-        /// <summary>
-        /// Get one task from an index.
-        /// </summary>
-        /// <param name="indexUid">Uid of the index.</param>
-        /// <param name="taskUid">Uid of the task.</param>
-        /// <param name="cancellationToken">The cancellation token for this call.</param>
-        /// <returns>Return the task of the index.</returns>
-        public async Task<TaskInfo> GetIndexTaskAsync(string indexUid, int taskUid, CancellationToken cancellationToken = default)
```
brunoocasali, line 59: If I'm correct, we agreed to keep those methods but remove the implementation that points to `$"indexes/{indexUid}/tasks/{taskUid}"` and replace it to `$"indexes/{taskUid}"`

src/Meilisearch/TaskInfo.cs
```diff
@@ -28,21 +28,21 @@ public class TaskInfo
...
 
         /// <summary>
         /// The status of the task. Possible values are enqueued, processing, succeeded, failed.
         /// </summary>
         public TaskInfoStatus Status { get; }
 
         /// <summary>
-        /// The type of task. Possible values are indexCreation, indexUpdate, indexDeletion, documentAddition,
-        /// documentPartial, documentDeletion, settingsUpdate, clearAll.
+        /// The type of task. Possible values are indexCreation, indexUpdate, indexDeletion,
```
brunoocasali: I strongly believe you can remove this comment because all the possible types are already described in line 90, and that line is the only that matters :)

src/Meilisearch/TasksResults.cs
```diff
@@ -0,0 +1,22 @@
+using System.Collections.Generic;
+
+namespace Meilisearch
+{
+    /// <summary>
+    /// Generic result class for ressources.
```
brunoocasali:
```suggestion
    /// Generic result class for resources.
```

src/Meilisearch/TaskRessource.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// </summary>
+        public string? IndexUid { get; }
+
+        /// <summary>
+        /// The status of the task. Possible values are enqueued, processing, succeeded, failed.
+        /// </summary>
+        public TaskInfoStatus Status { get; }
+
+        /// <summary>
+        /// The type of task. Possible values are indexCreation, indexUpdate, indexDeletion,
```
brunoocasali: The same idea as before, we can remove here :)

src/Meilisearch/TaskRessource.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// documentAdditionOrUpdate, documentDeletion, settingsUpdate.
+        /// </summary>
+        public TaskInfoType Type { get; }
+
+        /// <summary>
+        /// Detailed information on the task payload.
+        /// </summary>
+        public Dictionary<string, object> Details { get; }
```
brunoocasali: You can use `dynamic` if you don't know anything about the type :)

```suggestion
        public Dictionary<string, dynamic> Details { get; }
```

tests/Meilisearch.Tests/SettingsTests.cs
```diff
@@ -468,9 +468,9 @@ private async Task AssertGetInequality<TValue>(IndexGetMethod<TValue> getMethod,
 
         private async Task AssertTaskInfoSucceeded(TaskInfo task)
         {
-            task.Uid.Should().BeGreaterThan(0);
-            task = await _index.WaitForTaskAsync(task.Uid);
-            task.Status.Should().Be(TaskInfoStatus.Succeeded);
+            task.TaskUid.Should().BeGreaterThan(0);
+            var taskRessource = await _index.WaitForTaskAsync(task.TaskUid);
```
brunoocasali: Why not just `task`? I believe the only protection is when we call `Task` with a capital letter.
```suggestion
            var task = await _index.WaitForTaskAsync(task.TaskUid);
```

brunoocasali (changes requested):
Awesome work by creating two types to handle the tasks ;) 🎉 

I had some comments, I think you can address them!

src/Meilisearch/TaskEndpoint.cs
alallema: In the [Issue](https://github.com/meilisearch/integration-guides/issues/205) it says:

> All the changes:

>Remove GET /indexes/:indexUid/tasks. Use GET /tasks?indexUid=:indexUid instead.
Remove GET /indexes/:indexUid/tasks/:taskUid. Use GET /tasks/:taskUid instead.

So I just used the same `GetTaskAsync()` method for client and index

src/Meilisearch/TaskRessource.cs
alallema: Thanks I didn't know that

src/Meilisearch/Index.Tasks.cs
alallema: Actually it should be `TaskResults`

tests/Meilisearch.Tests/SettingsTests.cs
alallema: Because the function parameter is already called task. This is not a problem?

brunoocasali: yep, that's a big problem hahaha 😬

tests/Meilisearch.Tests/SettingsTests.cs
```diff
@@ -468,9 +468,9 @@ private async Task AssertGetInequality<TValue>(IndexGetMethod<TValue> getMethod,
 
         private async Task AssertTaskInfoSucceeded(TaskInfo task)
```
brunoocasali, line 469: `TaskInfo` does not exist anymore correct?

alallema: It does, I create to type to handle Task maybe it's not the best to do because of the heritage in c# but I had a type for the return of asynchronous methods and one for the return of `GetTask` and `GetTasks` who is returning e real Task resource

brunoocasali (approved):
🎉 🎉

PR #289, opened 2022-07-12
Apply indexes changes

alallema:
# Pull Request

Related to: https://github.com/meilisearch/meilisearch/issues/2373

### Changes

- Add `public class ResourceResults<T> : Result<T>`
- Wrap the indexes objects in a `RessourceResults` class.

### DIFF
--- a/src/Meilisearch/MeilisearchClient.cs
+++ b/src/Meilisearch/MeilisearchClient.cs
@@ -122,20 +122,12 @@ namespace Meilisearch
         /// </summary>
         /// <param name="cancellationToken">The cancellation token for this call.</param>
         /// <returns>An IEnumerable of indexes in JsonElement format.</returns>
-        public async Task<IEnumerable<JsonElement>> GetAllRawIndexesAsync(CancellationToken cancellationToken = default)
+        public async Task<JsonDocument> GetAllRawIndexesAsync(CancellationToken cancellationToken = default)
         {
             var response = await _http.GetAsync("indexes", cancellationToken).ConfigureAwait(false);
 
             var content = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
-            var json = JsonDocument.Parse(content);
-            var indexes = new List<JsonElement>();
-
-            foreach (var element in json.RootElement.EnumerateArray())
-            {
-                indexes.Add(element);
-            }
-
-            return indexes;
+            return JsonDocument.Parse(content);
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/Meilisearch/MeilisearchClient.cs
```diff
@@ -122,34 +122,27 @@ public async Task<TaskInfo> DeleteIndexAsync(string uid, CancellationToken cance
...
-            var json = JsonDocument.Parse(content);
-            var indexes = new List<JsonElement>();
-
-            foreach (var element in json.RootElement.EnumerateArray())
-            {
-                indexes.Add(element);
-            }
-
-            return indexes;
+            return JsonDocument.Parse(content);
```
brunoocasali, line 130: I love the simplification, but I would love more when we drop these raw methods 😬

src/Meilisearch/MeilisearchClient.cs
```diff
@@ -122,34 +122,27 @@ public async Task<TaskInfo> DeleteIndexAsync(string uid, CancellationToken cance
...
         /// <returns>Return Enumerable of Index.</returns>
-        public async Task<IEnumerable<Index>> GetAllIndexesAsync(CancellationToken cancellationToken = default)
+        public async Task<ResourceResults<IEnumerable<Index>>> GetAllIndexesAsync(CancellationToken cancellationToken = default)
         {
             var response = await _http.GetAsync("indexes", cancellationToken).ConfigureAwait(false);
 
-            var content = await response.Content.ReadFromJsonAsync<IEnumerable<Index>>(cancellationToken: cancellationToken).ConfigureAwait(false);
-            return content
+            var content = await response.Content.ReadFromJsonAsync<ResourceResults<IEnumerable<Index>>>(cancellationToken: cancellationToken).ConfigureAwait(false);
+            content.Results
```
brunoocasali, line 143: I'm not so sure about what this line 143 is doing... Can you check if it is important?

Because I assume the result from the LINQ query `.Select` will not be applied to the `content` variable.

tests/Meilisearch.Tests/IndexFixture.cs
```diff
@@ -197,9 +197,9 @@ public async Task<Index> SetUpIndexForNestedSearch(string indexUid)
         public async Task DeleteAllIndexes()
         {
             var indexes = await DefaultClient.GetAllIndexesAsync();
-            foreach (var index in indexes)
+            foreach (var index in indexes.Results)
```
brunoocasali, line 200: Are you returning an `Index` from this list, don't you?

So I think you don't need the `DefaultClient.Index(index.Uid)` you can just `index.DeleteAsync();`

src/Meilisearch/MeilisearchClient.cs
alallema: I know I hesitate to remove it too ...

src/Meilisearch/MeilisearchClient.cs
alallema: Yes, I had an issue where it should add the `HttpClient _http` but I'm not sure it works well ...

tests/Meilisearch.Tests/IndexFixture.cs
alallema: It's because line 143 is not working properly so it didn't work.

src/Meilisearch/MeilisearchClient.cs
brunoocasali: I think I got it, it adds the HTTP client as a field/attribute to the instance of that object, in this case, the index and the user can make requests normally without worrying about the `HttpClient`.

alallema: It should but It doesn't seem to work properly inside the code even if I try the function outside and it works fine

brunoocasali (approved):
🎉 🎉

PR #290, opened 2022-07-12
Apply documents changes

alallema:
# Pull Request

Related to: https://github.com/meilisearch/integration-guides/issues/205

## What does this PR do?

- `GET /documents/:uid`, `GET /documents` Add the possibility to reduce the body payload by using a query parameter called `fields` (previously called `attributesToRetrieve`), check the issue and the spec for the entire behavior.
- The documents objects in now return wrap in a `ResourceResults`

### DIFF
--- a/src/Meilisearch/DocumentQuery.cs
+++ b/src/Meilisearch/DocumentQuery.cs
@@ -1,3 +1,5 @@
+using System.Collections.Generic;
+
 namespace Meilisearch
 {
     /// <summary>
@@ -16,8 +18,8 @@ namespace Meilisearch
         public int? Limit { get; set; }
 
         /// <summary>
-        /// Gets or sets the attributes to retrieve..
+        /// Gets or sets the attributes to retrieve.
         /// </summary>
-        public IEnumerable<string> Fields { get; set; }
+        public List<string>? Fields { get; set; }
     }
 }
--- a/src/Meilisearch/DocumentQuery.cs
+++ b/src/Meilisearch/DocumentQuery.cs
@@ -16,8 +16,8 @@ namespace Meilisearch
         public int? Limit { get; set; }
 
         /// <summary>
-        /// Gets or sets the attributes to retrieve.
+        /// Gets or sets the attributes to retrieve..
         /// </summary>
-        public string AttributesToRetrieve { get; set; }
+        public IEnumerable<string> Fields { get; set; }
     }
 }
--- a/src/Meilisearch/Index.Documents.cs
+++ b/src/Meilisearch/Index.Documents.cs
@@ -329,10 +329,16 @@ namespace Meilisearch
         /// <param name="cancellationToken">The cancellation token for this call.</param>
         /// <typeparam name="T">Type of the document.</typeparam>
         /// <returns>Returns the document, with the according type if the object is available.</returns>
-        public async Task<T> GetDocumentAsync<T>(string documentId, CancellationToken cancellationToken = default)
+        public async Task<T> GetDocumentAsync<T>(string documentId, List<string>? fields = default, CancellationToken cancellationToken = default)
         {
+            var uri = $"indexes/{Uid}/documents/{documentId}";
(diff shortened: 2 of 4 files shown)


src/Meilisearch/DocumentQuery.cs
```diff
@@ -16,8 +16,8 @@ public class DocumentQuery
         public int? Limit { get; set; }
 
         /// <summary>
-        /// Gets or sets the attributes to retrieve.
+        /// Gets or sets the fields.
```
brunoocasali: I think you could even keep the old description it was clear :D

src/Meilisearch/Index.Documents.cs
```diff
@@ -355,7 +355,7 @@ public async Task<T> GetDocumentAsync<T>(int documentId, CancellationToken cance
         /// <param name="cancellationToken">The cancellation token for this call.</param>
         /// <typeparam name="T">Type of the document.</typeparam>
         /// <returns>Returns the list of documents.</returns>
-        public async Task<IEnumerable<T>> GetDocumentsAsync<T>(DocumentQuery query = default,
+        public async Task<ResourceResults<IEnumerable<T>>> GetDocumentsAsync<T>(DocumentQuery? query = default,
```
brunoocasali, line 364: In line 346 https://github.com/meilisearch/meilisearch-dotnet/pull/290/files#diff-b99cc3e20529ed19522ca88352dc425119bb704d9bf108d5dace1ef05b8e1941L346 you can add the `List<string> fields?` as an argument as well.

src/Meilisearch/DocumentQuery.cs
```diff
@@ -16,8 +16,8 @@ public class DocumentQuery
...
         /// </summary>
-        public string AttributesToRetrieve { get; set; }
+        public string Fields { get; set; }
```
brunoocasali: I strongly think this should be an array/list. As a user, it will be easier to send the data to the SDK if it will be an array, and if Meili changes the format of the separator, we can ensure it will be properly change :)

brunoocasali (changes requested):

src/Meilisearch/Index.Documents.cs
```diff
@@ -329,10 +329,16 @@ public partial class Index
         /// <param name="cancellationToken">The cancellation token for this call.</param>
         /// <typeparam name="T">Type of the document.</typeparam>
         /// <returns>Returns the document, with the according type if the object is available.</returns>
-        public async Task<T> GetDocumentAsync<T>(string documentId, CancellationToken cancellationToken = default)
+        public async Task<T> GetDocumentAsync<T>(string documentId, List<string>? fields = default, CancellationToken cancellationToken = default)
         {
+            var uri = $"indexes/{Uid}/documents/{documentId}";
+            if (query != null)
```
brunoocasali: Where did this comes from? 😬

tests/Meilisearch.Tests/DocumentTests.cs
```diff
@@ -516,19 +516,19 @@ public async Task GetMultipleExistingDocuments()
...
         public async Task GetMultipleExistingDocumentsWithLimit()
         {
             var index = await _fixture.SetUpBasicIndex("GetMultipleExistingDocumentWithLimitTest");
             var documents = await index.GetDocumentsAsync<Movie>(new DocumentQuery() { Limit = 2 });
-            Assert.Equal(2, documents.Count());
-            documents.First().Id.Should().Be("10");
-            documents.Last().Id.Should().Be("11");
+            Assert.Equal(2, documents.Results.Count());
+            documents.Results.First().Id.Should().Be("10");
+            documents.Results.Last().Id.Should().Be("11");
```
brunoocasali, line 551: I miss two tests:

- one for `GetDocumentAsync` with `fields`
- one for `GetDocumentsAsync` with `fields` in the `DocumentsQuery`

brunoocasali (approved):
🎉

PR #293, opened 2022-07-13
Apply changes for HTTP method patch

alallema:

Add last HTTP changes after the addition of `PATCH`
- `PATCH` `/indexes/{indexUid}` instead of `PUT`
- `PATCH` `/indexes/{indexUid}/settings` instead of `POST`
- `PATCH` `/indexes/{indexUid}/settings/typo-tolerance` instead of `POST`


brunoocasali (approved):
on fire! 🧯

PR #299, opened 2022-07-19
Remove nullable and force langage to 8

alallema:
When implementing the `PATCH` method the `nullable` option has been enabled, this forces to use version 8 of the language which would be a breaking change. This has been removed


brunoocasali (approved):
🎉 🎉

PR #300, opened 2022-07-25
Apply changes for paginations in multiples methods

alallema:
Add pagination filter to all these routes and methods:
- GET /tasks has pagination metadata, `type`, `status`, and `indexUid` in `AsyncGetTaks()`
- GET /keys has pagination metadata, added limit (default: 20), offset (default: 0), total in `AsyncGetKeys()`
- GET /indexes has pagination metadata, added limit (default: 20), offset (default: 0), total in `AsyncGetIndexes()`
- GET /documents has pagination metadata, added limit (default: 20), offset (default: 0), total in `AsyncGetDocuments()`

### DIFF
--- a/src/Meilisearch/Extensions/ObjectExtensions.cs
+++ b/src/Meilisearch/Extensions/ObjectExtensions.cs
@@ -24,41 +24,26 @@ namespace Meilisearch.Extensions
         }
 
         /// <summary>
-        /// Transforms a Meilisearch object into an URL encoded query string.
+        /// Transforms a Meilisearch object containing Lists into an URL encoded query string.
         /// </summary>
         /// <param name="source">Object to transform.</param>
         /// <param name="bindingAttr">Binding flags.</param>
         /// <returns>Returns an url encoded query string.</returns>
         internal static string ToQueryString(this object source, BindingFlags bindingAttr = BindingFlags.DeclaredOnly | BindingFlags.Public | BindingFlags.Instance)
         {
-            var values = source.GetType().GetProperties(bindingAttr)
-            .Where(p => p.GetValue(source, null) != null)
-            .Select(p =>
-                Uri.EscapeDataString(char.ToLowerInvariant(p.Name[0]) + p.Name.Substring(1)) + "=" + Uri.EscapeDataString(p.GetValue(source, null).ToString()));
-            var queryString = string.Join("&", values);
-            return queryString;
-        }
-
-        /// <summary>
-        /// Transforms a Meilisearch object containing Lists into an URL encoded query string.
-        /// </summary>
-        /// <param name="source">Object to transform.</param>
-        /// <param name="bindingAttr">Binding flags.</param>
-        /// <returns>Returns an url encoded query string.</returns>
-        internal static string ToQueryStringWithList(this object source, BindingFlags bindingAttr = BindingFlags.DeclaredOnly | BindingFlags.Public | BindingFlags.Instance)
-        {
-            var values = new List<string>(); ;
-            foreach (var p in source.GetType().GetProperties(bindingAttr))
+            var values = new List<string>();
+            foreach (var field in source.GetType().GetProperties(bindingAttr))
             {
-                if (p.GetValue(source, null) != null)
+                if (field.GetValue(source, null) != null)
                 {
-                    if (!(p.GetValue(source, null).GetType().IsGenericType && p.GetValue(source, null).GetType().GetGenericTypeDefinition() == typeof(List<>)))
+                    var isList = field.GetValue(source, null).GetType().IsGenericType && field.GetValue(source, null).GetType().GetGenericTypeDefinition() == typeof(List<>);
+                    if (isList)
                     {
-                        values.Add(Uri.EscapeDataString(char.ToLowerInvariant(p.Name[0]) + p.Name.Substring(1)) + "=" + Uri.EscapeDataString(p.GetValue(source, null).ToString()));
+                        values.Add(Uri.EscapeDataString(char.ToLowerInvariant(field.Name[0]) + field.Name.Substring(1)) + "=" + string.Join(",", (List<string>)field.GetValue(source, null)));
                     }
-                    if (p.GetValue(source, null).GetType().IsGenericType && p.GetValue(source, null).GetType().GetGenericTypeDefinition() == typeof(List<>))
+                    else
                     {
-                        values.Add(Uri.EscapeDataString(char.ToLowerInvariant(p.Name[0]) + p.Name.Substring(1)) + "=" + string.Join(",", (List<string>)p.GetValue(source, null)));
+                        values.Add(Uri.EscapeDataString(char.ToLowerInvariant(field.Name[0]) + field.Name.Substring(1)) + "=" + Uri.EscapeDataString(field.GetValue(source, null).ToString()));
                     }
                 }
             }
--- a/src/Meilisearch/Extensions/ObjectExtensions.cs
+++ b/src/Meilisearch/Extensions/ObjectExtensions.cs
@@ -38,5 +38,31 @@ namespace Meilisearch.Extensions
             var queryString = string.Join("&", values);
             return queryString;
         }
+
+        /// <summary>
+        /// Transforms a Meilisearch object containing Lists into an URL encoded query string.
+        /// </summary>
+        /// <param name="source">Object to transform.</param>
+        /// <param name="bindingAttr">Binding flags.</param>
+        /// <returns>Returns an url encoded query string.</returns>
+        internal static string ToQueryStringWithList(this object source, BindingFlags bindingAttr = BindingFlags.DeclaredOnly | BindingFlags.Public | BindingFlags.Instance)
+        {
+            var values = new List<string>(); ;
+            foreach (var p in source.GetType().GetProperties(bindingAttr))
+            {
+                if (p.GetValue(source, null) != null)
+                {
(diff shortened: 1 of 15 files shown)


tests/Meilisearch.Tests/ObjectExtensionsTests.cs
```diff
@@ -35,11 +37,47 @@ public void QueryStringsAreEqualsForPrimaryKey(string key)
...
             Assert.Equal(expected, actual);
         }
+
+        [Theory]
+        [InlineData(null, null, null)]
+        [InlineData(1, null, null)]
+        [InlineData(null, 3, null)]
+        [InlineData(null, null, new string[] { "attr" })]
+        [InlineData(null, null, new string[] { "attr", "attr2", "attr3" })]
+        [InlineData(1, 2, null)]
```
brunoocasali, line 34: Meilisearch actually supports `fields=` (without value), but the result is not nice to the user:

```json
{
    "results": [
        {},
        {},
        {},
        {},
        {},
        {}
    ],
    "offset": 0,
    "limit": 20,
    "total": 6
}
```

So I think we should add a validation to prevent null values being added in the final query.

src/Meilisearch/Extensions/ObjectExtensions.cs
```diff
@@ -38,5 +38,31 @@ internal static string ToQueryString(this object source, BindingFlags bindingAtt
...
             return queryString;
         }
+
+        /// <summary>
+        /// Transforms a Meilisearch object containing Lists into an URL encoded query string.
+        /// </summary>
+        /// <param name="source">Object to transform.</param>
+        /// <param name="bindingAttr">Binding flags.</param>
+        /// <returns>Returns an url encoded query string.</returns>
+        internal static string ToQueryStringWithList(this object source, BindingFlags bindingAttr = BindingFlags.DeclaredOnly | BindingFlags.Public | BindingFlags.Instance)
```
brunoocasali: I thought this was really complex to understand. So I took some time to really understand and suggest some changes:

- Remove the method `ToQueryString` in the line 32.
- Rename this method to `ToQueryString` and all of the occurrences to `ToQueryString`.
- `var p in source.GetType().GetProperties(bindingAttr)` is not easy to guess in the loop, so we could rename it to `field`, which is the name C# calls the class attributes. 
- Use an if/else inside of line 53 instead of two ifs because you just want to give different handling to the lists.
- I suggest defining a variable like this `var isList = field.GetValue(source, null).GetType().IsGenericType && field.GetValue(source, null).GetType().GetGenericTypeDefinition() == typeof(List<>);` to increase readability and use this variable in the if/else :)
- Instead of creating a `List<string>()` we can define a `Dictionary` + `System.Net.Http.FormUrlEncodedContent` to generate the query string.
- I think we should also prevent this extension from being used incorrectly by defining a more specific type instead of using `object`, like: `internal static string ToQueryString(this IQuery source`. And of course we have to create this interface 😬  
- If you do that, you'll have to change the tests because now all the results will be ascii encoded :)

src/Meilisearch/Index.Documents.cs
```diff
@@ -357,22 +358,17 @@ public async Task<T> GetDocumentAsync<T>(int documentId, List<string> fields = d
...
+        /// <param name="query">Query parameters. Supports limit, offset and attributes to retrieve named fields.</param>
         /// <param name="cancellationToken">The cancellation token for this call.</param>
         /// <typeparam name="T">Type of the document.</typeparam>
         /// <returns>Returns the list of documents.</returns>
-        public async Task<ResourceResults<IEnumerable<T>>> GetDocumentsAsync<T>(DocumentQuery query = default,
+        public async Task<ResourceResults<IEnumerable<T>>> GetDocumentsAsync<T>(DocumentsQuery query = default,
             CancellationToken cancellationToken = default)
         {
             var uri = $"indexes/{Uid}/documents";
             if (query != null)
```
brunoocasali, line 369: This condition is not good because we are doing that multiple times. What do you think to create an issue to refactor this?

My suggestion is to use the `Uri` object to handle that.

src/Meilisearch/QueryParamaters/DocumentsQuery.cs
```diff
@@ -1,21 +1,21 @@
 using System.Collections.Generic;
 
-namespace Meilisearch
+namespace Meilisearch.QueryParameters
 {
     /// <summary>
-    /// Document Query for Meilisearch class.
+    /// Documents Query for Meilisearch class.
```
brunoocasali: What do you think?
```suggestion
    /// A class that handles the creation of a query string for Documents.
```

src/Meilisearch/TaskEndpoint.cs
```diff
@@ -17,11 +21,17 @@ public class TaskEndpoint
         /// <summary>
         /// Gets the tasks.
         /// </summary>
+        /// <param name="query">Query parameters. Supports limit, from, indexUid, status and types.</param>
```
brunoocasali: I like this documentation here, but I think this could be outdated if we forgot to add a new value here 😬

src/Meilisearch/QueryParamaters/DocumentsQuery.cs
brunoocasali: If you agree you can apply these to the Keys, Indexes, Tasks as well 😊

tests/Meilisearch.Tests/TaskInfoTests.cs
```diff
@@ -26,14 +32,42 @@ public async Task InitializeAsync()
...
         }
 
+        [Fact]
+        public async Task GetMultipleTaskInfoWithLimitFromClient()
+        {
+            await _index.AddDocumentsAsync(new[] { new Movie { Id = "1" } });
+            var taskResponse = await _client.GetTasksAsync(new TasksQuery { Limit = 1 });
+            var tasks = taskResponse;
+            tasks.Results.Count().Should().BeGreaterOrEqualTo(1);
+            Assert.Equal(1, tasks.Limit);
```
brunoocasali:
```suggestion
            await _index.AddDocumentsAsync(new[] { new Movie { Id = "1" } });
            var taskResponse = await _client.GetTasksAsync(new TasksQuery { Limit = 1 });

            taskResponse.Results.Count().Should().BeGreaterOrEqualTo(1);
            Assert.Equal(1, tasks.Limit);
```

tests/Meilisearch.Tests/TaskInfoTests.cs
```diff
@@ -26,14 +32,42 @@ public async Task InitializeAsync()
...
+        }
+
+        [Fact]
+        public async Task GetMultipleTaskInfoWithQueryParameters()
+        {
+            await _index.AddDocumentsAsync(new[] { new Movie { Id = "1" } });
+            var taskResponse = await _index.GetTasksAsync(new TasksQuery { Limit = 1, IndexUid = new List<string> { _index.Uid } });
+            var tasks = taskResponse.Results;
+            tasks.Count().Should().BeGreaterOrEqualTo(1);
```
brunoocasali:
```suggestion
            await _index.AddDocumentsAsync(new[] { new Movie { Id = "1" } });
            var taskResponse = await _index.GetTasksAsync(new TasksQuery { Limit = 1, IndexUid = new List<string> { _index.Uid } });

            taskResponse.Results.Count().Should().BeGreaterOrEqualTo(1);
```

brunoocasali (changes requested):
Hi @alallema, this is good! Congratulations!

I thought of some ways to improve the simplicity of the code. Take a look, please! :)

src/Meilisearch/Extensions/ObjectExtensions.cs
alallema: > Instead of creating a List<string>() we can define a Dictionary + System.Net.Http.FormUrlEncodedContent to generate the query string.

I'm not sure to understand what you mean?

>I think we should also prevent this extension from being used incorrectly by defining a more specific type instead of using object, like: internal static string ToQueryString(this IQuery source. And of course we have to create this interface 😬
If you do that, you'll have to change the tests because now all the results will be ascii encoded :)

I think that this one could be dealt with later and for the moment creating an issue will end up not?

src/Meilisearch/Index.Documents.cs
alallema: You mean to create a method to handle `if (query != null)` like 
```
    var uri = computeUri(String uri, DocumentQuery query)
```
?

brunoocasali: Not exactly this, but if we can just append the query to the URI without having to check all the time, will be a great addition.

If we use something like the uribuilder could help :)

https://docs.microsoft.com/en-us/dotnet/api/system.uribuilder.query?view=net-6.0

tests/Meilisearch.Tests/ObjectExtensionsTests.cs
alallema: I agree with you about the result but is it up to the SDK to handle this if Meilisearch does handle it? I'm not sure

src/Meilisearch/Index.Documents.cs
alallema: But they seem to be doing the same check as us no?
```
if (baseUri.Query != null && baseUri.Query.Length > 1)
    baseUri.Query = baseUri.Query.Substring(1) + "&" + queryToAppend; 
```

tests/Meilisearch.Tests/ObjectExtensionsTests.cs
brunoocasali: This is just a nice addition, unless we can think in a use case that behavior is required...

brunoocasali (approved):
🎉🍱🥓 LGTM!!

PR #302, opened 2022-07-28
Misc changes related to v0.28

brunoocasali:
- Fixed some definitions in the code-samples (+ try to reduce complexity from some samples).
- Add the automatic filter by `indexUid` if the user calls `index.GetKeysAsync`.
- Add a use-case to cover optional `uid` in the KeysTests.


brunoocasali:
@alallema, as we aligned internally, I'll merge this, but I would like to have your feedback in any case.

PR #330, opened 2022-10-06
Change SearchAsync<T> signature

juchom:
# Pull Request

## What does this PR do?
The old signature was a bit akward, it took as first paramater a string for the query and a second parameter an optional SearchQuery object where we store later the first parameter into the Q property of the SearchQuery.

```csharp
// Old signature
public async Task<SearchResult<T>> SearchAsync<T>(string query,
            SearchQuery searchAttributes = default(SearchQuery), CancellationToken cancellationToken = default)
{
}
```

The main issue was that the SearchQuery can be created with the default constructor which is wrong. The SearchQuery object must be created by passing a search query to the constructor to be in a valid state.

The new signature is now always using the SearchQuery parameter like in other SDK

```csharp
// New signature
public async Task<SearchResult<T>> SearchAsync<T>(SearchQuery searchQuery, CancellationToken cancellationToken = default)
{
}
```

## PR checklist
- [X] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -43,7 +43,7 @@ delete_one_document_1: |-
 delete_documents_1: |-
   await client.Index("movies").DeleteDocumentsAsync(new[] { "23488", "153738", "437035", "363869" });
 search_post_1: |-
-  await client.Index("movies").SearchAsync<Movie>("American ninja");
+  await client.Index("movies").SearchAsync<Movie>(new SearchQuery("American ninja"));
 get_task_1: |
   TaskInfo task = await client.GetTaskAsync(1);
 get_all_tasks_1: |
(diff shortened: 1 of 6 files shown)


alallema:
Hi @juchom,
Thanks a lot for your RP but the old signature although akward indeed is the right one. For example in other SDK it's:
In `php`:
```php
    public function search(?string $query, array $searchParams = [], array $options = [])
```
In `python`:
```python
    def search(self, query: str, opt_params: dict[str, Any] | None = None) -> dict[str, Any]
```
In `Go`:
```golang
    func (i Index) Search(query string, request *SearchRequest) (*SearchResponse, error)
```
It was decided by the integration team, that the signature always proposes the first parameter to be the query as `string`.
So I think maybe the issue is more how the `SearchQuery` is built after that and also I don't think it was a good naming `SearchQuery` is should be something like `SearchParam` or `SearchRequest` maybe.
I can see that your solution brings clarity to the code but we try as much as possible to keep a certain consistency between the different SDKs

juchom:
Sorry for the mistake, when I worked on the samples, I was pretty sure Java and Rust were using it like in this PR.

I understand what you want to do. You are right the `SearchQuery` object has a bad name it should be `SearchParameters` and should not expose a `Q` property.

I close this PR, because it's not in line with other SDKs. I will have a look later at this.

Thanks for the review 👍

alallema:
Yes I'm sorry but thank you very much for this PR anyway ❤️ 
Also Java is a bit of a special case

alallema:
_This message is sent automatically_

Thanks again for contributing to Meilisearch :heart:
If you are participating in [Hacktoberfest](https://blog.meilisearch.com/contribute-hacktoberfest-2022/), and you would like to receive some gift from Meilisearch too, please complete [this form](https://meilisearch.typeform.com/to/idt5qugY).

PR #351, opened 2022-10-21
Allow multiple pagination styles co-exist in the same response

brunoocasali:
This is part of a proof-of-concept that will be part of the next release, v0.30.

The idea of this PR is to accommodate in a strongly typed language the ability to handle two different response types: the estimation-based response and the finite pagination based.

More details here: https://github.com/meilisearch/meilisearch/discussions/2635.

For reviewers:

- [ ] Check if the ergonomics of the API are good enough.
- [ ] If possible try-it-out with [v0.30.0-pagination.beta.4](https://hub.docker.com/layers/getmeili/meilisearch/v0.30.0-pagination.beta.4/images/sha256-0c5207423dded2489a53ae8325b803d10f5313e08d712cfb5102eee9644dd41e?context=explore) add it to `.env`.

How to use:

```c#
// When you want to search with finite pagination:
await client.SearchAsync<Movie>("man", new SearchQuery { Page = 2 });

// When you want to search with finite pagination but also need additional data like (total, page, and hits per page):
var movies = (PaginatedSearchResult<Movie>) await client.SearchAsync<Movie>("man", new SearchQuery { Page = 2 });
var currentPage = movies.Page; 

// When you want to use the old-style estimated value:
await client.SearchAsync<Movie>("man", new SearchQuery { Limit = 20 });
await client.SearchAsync<Movie>("man", new SearchQuery { }); // just don't use HitsPerPage and Page.
```

### DIFF
--- a/src/Meilisearch/MatchPosition.cs
+++ b/src/Meilisearch/MatchPosition.cs
@@ -1,0 +1,27 @@
+using System.Text.Json.Serialization;
+
+namespace Meilisearch
+{
+    public class MatchPosition
+    {
+        public MatchPosition(int start, int length)
+        {
+            Start = start;
+            Length = length;
+        }
+
+        /// <summary>
+        /// The beginning of a matching term within a field.
+        /// WARNING: This value is in bytes and not the number of characters. For example, ü represents two bytes but one character.
+        /// </summary>
+        [JsonPropertyName("start")]
+        public int Start { get; }
+
+        /// <summary>
+        /// The length of a matching term within a field.
+        /// WARNING: This value is in bytes and not the number of characters. For example, ü represents two bytes but one character.
+        /// </summary>
+        [JsonPropertyName("length")]
+        public int Length { get; }
+    }
+}
--- a/src/Meilisearch/SearchResult.cs
+++ b/src/Meilisearch/SearchResult.cs
@@ -72,27 +72,4 @@ namespace Meilisearch
         [JsonPropertyName("_matchesPosition")]
         public IReadOnlyDictionary<string, IReadOnlyCollection<MatchPosition>> MatchesPostion { get; }
     }
-
-    public class MatchPosition
-    {
-        public MatchPosition(int start, int length)
-        {
-            Start = start;
-            Length = length;
-        }
-
-        /// <summary>
-        /// The beginning of a matching term within a field.
-        /// WARNING: This value is in bytes and not the number of characters. For example, ü represents two bytes but one character.
-        /// </summary>
-        [JsonPropertyName("start")]
-        public int Start { get; }
-
-        /// <summary>
-        /// The length of a matching term within a field.
-        /// WARNING: This value is in bytes and not the number of characters. For example, ü represents two bytes but one character.
-        /// </summary>
-        [JsonPropertyName("length")]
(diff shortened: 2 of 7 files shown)


alallema (commented):
Thanks for this PoC @brunoocasali!
This is a good thing for me and a good start to see how it can be implemented ✨

curquiza:
I'm sorry I will not have the time to review I guess :(

brunoocasali:
> I'm sorry I will not have the time to review I guess :(

Don't worry @curquiza it's ok :)

tests/Meilisearch.Tests/SearchTests.cs
```diff
@@ -72,6 +72,28 @@ public async Task CustomSearchWithLimit()
...
+            movies.Hits.First().Name.Should().NotBeEmpty();
+            movies.Hits.First().Genre.Should().NotBeEmpty();
+        }
+
+        [Fact]
+        public async Task CustomSearchWithPageWithoutTypeCast()
+        {
+            var movies = await _basicIndex.SearchAsync<Movie>("man", new SearchQuery { Page = 1 });
+
+            movies.Hits.First().Id.Should().NotBeEmpty();
```
alallema, line 92: Why we didn't check if `movies.Page` exist in this test?

src/Meilisearch/SearchResult.cs
```diff
@@ -7,7 +7,7 @@ namespace Meilisearch
     /// Wrapper for Search Results.
     /// </summary>
     /// <typeparam name="T">Hit type.</typeparam>
-    public class SearchResult<T>
+    public class SearchResult<T> : ISearchable<T>
```
alallema, line 10: How did you choose the name `ISearchable`? Just to know I'm really bad at naming

src/Meilisearch/MatchPosition.cs
```diff
@@ -0,0 +1,27 @@
+using System.Text.Json.Serialization;
```
alallema, line 1: Great to move it!

alallema (approved):
Great job 🚀 
LGTM! 🔥

src/Meilisearch/SearchResult.cs
brunoocasali: https://en.wikibooks.org/wiki/C_Sharp_Programming/Interfaces#:~:text=A%20CONVENTION%20used%20in%20the,()%20in%20the%20above%20example.

> A CONVENTION used in the .NET Framework (and likewise by many C# programmers) is to place an "I" at the beginning of an interface name to distinguish it from a class name. Another common interface naming convention is used when an interface declares only one key method, such as Draw() in the above example. The interface name is then formed as an adjective by adding the "...able" suffix. So, the interface name above could also be IDrawable. This convention is used throughout the .NET Framework.

tests/Meilisearch.Tests/SearchTests.cs
brunoocasali: It is because it does not exist. If you don't typecast, all the specific fields from the `PaginatedSearchResult` type will not be available. 

It means if you don't want to typecast, it is ok. Your request will still work. The only thing is that some fields will not be available for you.

PR #367, opened 2022-12-02
Feature/Add cancel tasks

brunoocasali:
- Create `CancelTasksQuery` param config
- Add `CancelTasksAsync` method to cancel tasks.

### DIFF
--- a/src/Meilisearch/QueryParameters/CancelTasksQuery.cs
+++ b/src/Meilisearch/QueryParameters/CancelTasksQuery.cs
@@ -26,7 +26,7 @@ namespace Meilisearch.QueryParameters
         /// <summary>
         /// Gets or sets the list of uids to filter on. Case-sensitive.
         /// </summary>
-        public List<string> Uids { get; set; }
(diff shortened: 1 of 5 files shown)


src/Meilisearch/QueryParameters/CancelTasksQuery.cs
```diff
@@ -0,0 +1,66 @@
+using System;
```
alallema, line 1: Why don't use the `TaskQuery` directly?

alallema (commented):
Great 🎉 
Just a question

src/Meilisearch/QueryParameters/CancelTasksQuery.cs
brunoocasali: Because in the future if we want to deprecate any of the filters available in the `Tasks` or in `CancelTasks` we will have a breaking change.

brunoocasali: So it is good if they can evolve individually!

PR #369, opened 2022-12-02
Enhancement/Object extensions

brunoocasali:
- Improve `ToQueryString()` extension method 
- Add support to `DateTime` parameters. They will be converted to RFC 3339.

### DIFF
--- a/tests/Meilisearch.Tests/ObjectExtensionsTests.cs
+++ b/tests/Meilisearch.Tests/ObjectExtensionsTests.cs
@@ -10,8 +10,43 @@ using Xunit;
 
 namespace Meilisearch.Tests
 {
+    public class FakeQuery
+    {
+        public DateTime? FakeDate { get; set; }
+        public string FakeString { get; set; }
+        public int? FakeInteger { get; set; }
+        public List<string> FakeStringList { get; set; }
+    }
+
     public class ObjectExtensionsTests
     {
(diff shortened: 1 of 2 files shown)


tests/Meilisearch.Tests/ObjectExtensionsTests.cs
```diff
@@ -10,8 +10,43 @@
...
 namespace Meilisearch.Tests
 {
+    public class FakeQuery
+    {
+        public DateTime? FakeDate { get; set; }
+        public string FakeString { get; set; }
+        public int? FakeInteger { get; set; }
+        public List<string> FakeStringList { get; set; }
+    }
+
```
alallema, line 20: Why call every variable `Fake` instead of `mock` or `test` for example?

alallema (commented):
Again some questions

tests/Meilisearch.Tests/ObjectExtensionsTests.cs
brunoocasali: It was just a preference, but the main reason was to avoid this `public string String { get; set; }`, which will "conflict" with the class name `String`. So to make all of them look the same, I opted to duplicate the class name.

Also, `Mock` is not good because those fields are not mocks at all :)
And `Test` could be a good choice! But again just preference 👍

PR #372, opened 2022-12-03
Feature/Add delete tasks

brunoocasali:
- Add `KeyAction`.`TasksDelete` enum option.
- Add `TaskInfoType`.`TaskDeletion` enum option.
- Add `DeleteTasksQuery` query string configuration class.
- Add `DeleteTasksAsync` method on `MeilisearchClient`
- Fix documentation in the `CancelTasksAsync()`

### DIFF
--- a/src/Meilisearch/MeilisearchClient.cs
+++ b/src/Meilisearch/MeilisearchClient.cs
@@ -324,10 +324,21 @@ namespace Meilisearch
         /// </summary>
         /// <param name="query">Query parameters supports by the method.</param>
         /// <param name="cancellationToken">The cancellation token for this call.</param>
-        /// <returns>Returns the created API key.</returns>
+        /// <returns>Returns the task info of finished task.</returns>
         public async Task<TaskInfo> CancelTasksAsync(CancelTasksQuery query, CancellationToken cancellationToken = default)
         {
             return await TaskEndpoint().CancelTasksAsync(query, cancellationToken).ConfigureAwait(false);
+        }
+
+        /// <summary>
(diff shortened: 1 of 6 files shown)


src/Meilisearch/QueryParameters/DeleteTasksQuery.cs
```diff
@@ -0,0 +1,66 @@
+using System;
```
alallema, line 1: Same question than for `CancelTasksQuery` why we don't use the same structure?

src/Meilisearch/MeilisearchClient.cs
```diff
@@ -324,12 +324,23 @@ await _http.PostAsJsonAsync("keys", keyOptions, Constants.JsonSerializerOptionsW
         /// </summary>
         /// <param name="query">Query parameters supports by the method.</param>
         /// <param name="cancellationToken">The cancellation token for this call.</param>
-        /// <returns>Returns the created API key.</returns>
+        /// <returns>Returns the task info of finished task.</returns>
```
alallema, line 327: Good catch!

alallema (commented):
Just few question again 😄

src/Meilisearch/QueryParameters/DeleteTasksQuery.cs
brunoocasali: It is the same reason, because in the future if we want to deprecate any of the filters available in the `TasksQuery` or in `DeleteTasksQuery` we will have a breaking change.

PR #373, opened 2022-12-03
Feature/Add Swap Indexes

brunoocasali:
- Create `SwapIndexesAsync`
- Add `TaskInfoType.IndexSwap` enum option

### DIFF
--- a/tests/Meilisearch.Tests/MeilisearchClientTests.cs
+++ b/tests/Meilisearch.Tests/MeilisearchClientTests.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Net.Http;
+using System.Text.Json;
 using System.Threading.Tasks;
 
 using FluentAssertions;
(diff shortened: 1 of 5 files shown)


tests/Meilisearch.Tests/MeilisearchClientTests.cs
```diff
@@ -126,6 +127,19 @@ public async Task DeleteTasks()
...
+        [Fact]
+        public async Task SwapIndexes()
+        {
+            var swaps = new List<IndexSwap> { new IndexSwap("indexA", "indexB") };
+            var response = await _defaultClient.SwapIndexesAsync(swaps);
+            var task = await _defaultClient.WaitForTaskAsync(response.TaskUid);
+
+            response.TaskUid.Should().Be(task.Uid);
+            response.Type.Should().Be(TaskInfoType.IndexSwap);
+            task.Status.Should().Be(TaskInfoStatus.Failed);
```
alallema, line 139: Why make a request that will fail?

alallema (commented):
Just one question

tests/Meilisearch.Tests/MeilisearchClientTests.cs
brunoocasali: Because, in this case, I'm not interested in seeing if the indexes' swap will work. It is not up to this SDK to test this.

So, since we don't have a mock structure yet in this SDK, I can easily send a failing request and then assert all the methods I want, like which type of task I receive, if the Details contain the values I want, and so on... It will be almost the same if we call a successful request!

PR #374, opened 2022-12-03
Update version for the next release (v0.12.2)

brunoocasali:
This version makes this package compatible with Meilisearch v0.30.0 :tada:
Check out the changelog of [Meilisearch v0.30.0](https://github.com/meilisearch/meilisearch/releases/tag/v0.30.0) for more information on the changes.

## 🚀 Enhancements

- Add `MeilisearchClient.CancelTasksAsync(CancelTasksQuery query)` (#367) @brunoocasali 
  - `CancelTasksQuery`:
    - `List<string> IndexUids`
    - `List<TaskInfoStatus> Statuses`
    - `List<TaskInfoType> Types`
    - `List<int> Uids`
    - `List<string> CanceledBy`
    - `DateTime? BeforeEnqueuedAt`
    - `DateTime? AfterEnqueuedAt`
    - `DateTime? BeforeStartedAt`
    - `DateTime? AfterStartedAt`
    - `DateTime? BeforeFinishedAt`
    - `DateTime? AfterFinishedAt`
- Add `MeilisearchClient.SwapIndexesAsync(List<IndexSwap> indexes)` (#373) @brunoocasali 
- Add `MeilisearchClient.DeleteTasksAsync(DeleteTasksQuery query)` (#372) @brunoocasali
  - `DeleteTasksQuery`:
    - `List<string> IndexUids`
    - `List<TaskInfoStatus> Statuses`
    - `List<TaskInfoType> Types`
    - `List<int> Uids`
    - `List<string> CanceledBy`
    - `DateTime? BeforeEnqueuedAt`
    - `DateTime? AfterEnqueuedAt`
    - `DateTime? BeforeStartedAt`
    - `DateTime? AfterStartedAt`
    - `DateTime? BeforeFinishedAt`
    - `DateTime? AfterFinishedAt`

- `TasksQuery` new filters:
  - `List<int> Uids`
  - `List<string> CanceledBy`
  - `DateTime? BeforeEnqueuedAt`
  - `DateTime? AfterEnqueuedAt`
  - `DateTime? BeforeStartedAt`
  - `DateTime? AfterStartedAt`
  - `DateTime? BeforeFinishedAt`
  - `DateTime? AfterFinishedAt`

New enum types:

- `TaskInfoType`:
  - `TaskDeletion`
  - `IndexSwap`
  - `TaskCancellation`
  - `SnapshotCreation`

- `KeyAction`:
  - `TasksDelete`
  - `TasksCancel`

- `TaskInfoStatus`:
[...]

PR #365, opened 2022-11-30
Rename TaskQuery fields (IndexUids & Statuses)

brunoocasali:
Rename `TaskInfoQuery` fields to match the v0.30 spec.
- from `IndexUid` to `IndexUids`.
- from `Status` to `Statuses`.

PR #376, opened 2022-12-05
Update version for the next release (v0.13.0)

brunoocasali:
When I created PR #374, I did not remember I had fixed a break in the #365


brunoocasali:
bors merge

PR #402, opened 2023-03-15
Add TotalHits and TotalPages fields to PaginatedSearchResult

### LINKED ISSUE #400: Error on fields of PaginatedSearchResult.cs
**Description**
There is an error on fields of PaginatedSearchResults. 

**Expected behavior**
There is a field called "total" that is not correct, and there are two fields that are missing: totalPages and totalHints. This is important for pagination with page numbers, because they are the exact number of results and pages.

brunoocasali:
Fixes https://github.com/meilisearch/meilisearch-dotnet/issues/400

You can use this as the release CHANGELOG @alallema: 
 
- Remove `Total` from `PaginatedSearchResult`.
- Add `TotalHits` and `TotalPages` to `PaginatedSearchResult`.

### DIFF
--- a/src/Meilisearch/PaginatedSearchResult.cs
+++ b/src/Meilisearch/PaginatedSearchResult.cs
@@ -9,15 +9,23 @@ namespace Meilisearch
     /// <typeparam name="T">Hit type.</typeparam>
     public class PaginatedSearchResult<T> : ISearchable<T>
     {
-        public PaginatedSearchResult(IReadOnlyCollection<T> hits, int hitsPerPage, int page, int total,
+        public PaginatedSearchResult(
+            IReadOnlyCollection<T> hits,
+            int hitsPerPage,
+            int page,
+            int totalHits,
+            int totalPages,
             IReadOnlyDictionary<string, IReadOnlyDictionary<string, int>> facetDistribution,
-            int processingTimeMs, string query,
-            IReadOnlyDictionary<string, IReadOnlyCollection<MatchPosition>> matchesPostion)
+            int processingTimeMs,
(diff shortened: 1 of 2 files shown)


src/Meilisearch/PaginatedSearchResult.cs
```diff
@@ -9,15 +9,23 @@ namespace Meilisearch
     /// <typeparam name="T">Hit type.</typeparam>
     public class PaginatedSearchResult<T> : ISearchable<T>
     {
-        public PaginatedSearchResult(IReadOnlyCollection<T> hits, int hitsPerPage, int page, int total,
+        public PaginatedSearchResult(
```
brunoocasali, line 12: I reorganized the attributes here (one in each line). But I also added `int totalPages` and replaced `int total` with `int totalHits`.

I'm unsure how this can affect the user because it does not seem to be used (I think the json handler takes care of that).

alallema (approved):
Thanks @brunoocasali ❤️ 
LGTM! 🌮

src/Meilisearch/PaginatedSearchResult.cs
alallema: I'm not sure to understand, Do you mean the json handler should transform `int total` by `totalHits`?

alallema:
bors merge

alallema:
bors merge

src/Meilisearch/PaginatedSearchResult.cs
brunoocasali: Yes, because the constructor does not seem to be used anywhere.

PR #405, opened 2023-03-28
use JsonContent instead of StringContent

### LINKED ISSUE #394: Use JsonContent instead of StringContent in HttpExtensions.PrepareJsonPayload
**Description**
Memory usage and performance would improve if this library used JsonContent instead of StringContent when serializing to JSON for use with HttpClient. This is because JsonContent serializes directly to UTF-8 bytes which are written directly to the HttpClient output stream without storing the entire document in memory at once. Using StringContent forces the object to be serialized to a UTF-16 string in memory, which will need to be garbage-collected later, and then forces the UTF-16 string to be converted to UTF-8 bytes before it can be written to the HttpClient output stream.

**Basic example**

```csharp
private static StringContent PrepareJsonPayload<T>(T body, JsonSerializerOptions options = null)
{
    options ??= Constants.JsonSerializerOptionsWriteNulls;
    var payload = JsonContent.Create(body, new MediaTypeHeaderValue("application/json"), options);
    payload.Headers.ContentType.CharSet = string.Empty;

    return payload;
}
```

ahmednfwela:
# Pull Request

## Related issue
Fixes #394

## What does this PR do?
- Uses `JsonContent` instead of `StringContent`

## PR checklist

### DIFF
--- a/src/Meilisearch/Extensions/HttpExtensions.cs
+++ b/src/Meilisearch/Extensions/HttpExtensions.cs
@@ -86,7 +86,7 @@ namespace Meilisearch.Extensions
         private static JsonContent PrepareJsonPayload<T>(T body, JsonSerializerOptions options = null)
         {
             options = options ?? Constants.JsonSerializerOptionsWriteNulls;
-            var payload = JsonContent.Create(body, new MediaTypeHeaderValue("application/json"), options);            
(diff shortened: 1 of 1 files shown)


alallema (approved):
Thanks, @ahmednfwela  for this PR ❤️ 
LGTM! 🔥

alallema:
bors merge

PR #406, opened 2023-03-28
make _formatted value user friendly :)

### LINKED ISSUE #315: Using _Formatted value when searching is not user friendly
While I was working on the code I saw something that is not user friendly for C# and Java developers.

If you look at the documentation for Highlighting : https://docs.meilisearch.com/reference/api/search.html#attributes-to-highlight

Here is the json response of the server :

```json
{
  "id": 50393,
  "title": "Kung Fu Panda Holiday",
  "poster": "https://image.tmdb.org/t/p/w1280/gp18R42TbSUlw9VnXFqyecm52lq.jpg",
  "overview": "The Winter Feast is Po's favorite holiday. Every year he and his father hang decorations, cook together, and serve noodle soup to the villagers. But this year Shifu informs Po that as Dragon Warrior, it is his duty to host the formal Winter Feast at the Jade Palace. Po is caught between his obligations as the Dragon Warrior and his family traditions: between Shifu and Mr. Ping.",
  "release_date": 1290729600,
  "_formatted": {
    "id": 50393,
    "title": "Kung Fu Panda Holiday",
    "poster": "https://image.tmdb.org/t/p/w1280/gp18R42TbSUlw9VnXFqyecm52lq.jpg",
    "overview": "The <em>Winter Feast</em> is Po's favorite holiday. Every year he and his father hang decorations, cook together, and serve noodle soup to the villagers. But this year Shifu informs Po that as Dragon Warrior, it is his duty to host the formal <em>Winter Feast</em> at the Jade Palace. Po is caught between his obligations as the Dragon Warrior and his family traditions: between Shifu and Mr. Ping.",
    "release_date": 1290729600
  }
}
```

The server returns the document and add an extra field `_formatted` with a modified copy of the document.

Let see the problem now with C#:

We're going to work with this unit test :

https://github.com/meilisearch/meilisearch-dotnet/blob/3302d2b7c529c36de31298aaed7ac62f414170a1/tests/Meilisearch.Tests/SearchTests.cs#L76-L94

We are working on the `_basicIndex` and here is the definition of it :

https://github.com/meilisearch/meilisearch-dotnet/blob/3302d2b7c529c36de31298aaed7ac62f414170a1/tests/Meilisearch.Tests/IndexFixture.cs#L46-L69

We are adding a collection of `Movie` and here is the definition :

https://github.com/meilisearch/meilisearch-dotnet/blob/3302d2b7c529c36de31298aaed7ac62f414170a1/tests/Meilisearch.Tests/Movie.cs#L3-L10

Now, on line 86 of the test we are doing a Search query against Meilisearch with this code :

```csharp
     var movies = await _basicIndex.SearchAsync<FormattedMovie>( 
         "man", 
         new SearchQuery { AttributesToHighlight = new string[] { "name" } }); 
```

We are querying an index that contains `Movie` documents, but we ask to map to another class `FormattedMovie` with this definition :

https://github.com/meilisearch/meilisearch-dotnet/blob/3302d2b7c529c36de31298aaed7ac62f414170a1/tests/Meilisearch.Tests/Movie.cs#L38-L48

We have to duplicate the `Movie` class with a `FormattedMovie` class in order to have the `_Formatted` property available.

This is not very user friendly, because users will have to create two objects in order to fully use Meilisearch.

There is no way in C# to create a specific object that will enhance a base object with a `_Formatted` property like this :

```csharp
public class FormattedDocument<T> : T
{
    public T _Formatted { get; set; }
}
```

There is actually also a small risk of collision if a document contains a `_formatted` property.

I told Guillaume it would great to have a response with an `original` and `formatted` properties that contains both a version of the document.

```json
{
   "original":{
      "id":50393,
      "title":"Kung Fu Panda Holiday",
      "poster":"https://image.tmdb.org/t/p/w1280/gp18R42TbSUlw9VnXFqyecm52lq.jpg",
      "overview":"The Winter Feast is Po's favorite holiday. Every year he and his father hang decorations, cook together, and serve noodle soup to the villagers. But this year Shifu informs Po that as Dragon Warrior, it is his duty to host the formal Winter Feast at the Jade Palace. Po is caught between his obligations as the Dragon Warrior and his family traditions: between Shifu and Mr. Ping.",
      "release_date":1290729600
   },
   "formatted":{
      "id":50393,
      "title":"Kung Fu Panda Holiday",
      "poster":"https://image.tmdb.org/t/p/w1280/gp18R42TbSUlw9VnXFqyecm52lq.jpg",
      "overview":"The <em>Winter Feast</em> is Po's favorite holiday. Every year he and his father hang decorations, cook together, and serve noodle soup to the villagers. But this year Shifu informs Po that as Dragon Warrior, it is his duty to host the formal <em>Winter Feast</em> at the Jade Palace. Po is caught between his obligations as the Dragon Warrior and his family traditions: between Shifu and Mr. Ping.",
      "release_date":1290729600
   }
}
```

One more interesting thing in having a response like the one above is that if you want to add extra information like the global score of this result, or a score per field, you could enhance it easily without any risk of fields collisions.

But it's too late for breaking changes. So we need to find a way for developers (C# / Java) to enjoy this feature without too much pain.

Ping @gmourier @brunoocasali

ahmednfwela:
# Pull Request

## Related issue
Fixes #315

## What does this PR do?
- Introduces a new type `IFormatContainer<TOriginal, TFormatted>` coupled with a json converter `IFormatContainerJsonConverter<TOriginal, TFormatted>` and a `IFormatContainerJsonConverterFactory`
- Introduces a new `SearchAsync` overload 
  ```csharp
  public async Task<ISearchable<IFormatContainer<T, TFormatted>>> SearchAsync<T, TFormatted>(string query,
            SearchQuery searchAttributes = default, CancellationToken cancellationToken = default)
  ```
  to support `_formatted` documents.
- Modified tests to remove the ugly looking `FormattedMovie` in favor of the new API
- No breaking changes :)

Usage is simple:

- If the model has same signature as the "_formatted" model: use `SearchAsync<Model, Model>`
- If not, use `SearchAsync<Model, FormattedModel>`

## PR checklist

cc @alallema


juchom:
Hello @ahmednfwela,

Thanks for this PR and this idea to solve the issue.

The main question we have here, is to consider if the issue is coming from the client or from the server response.

My personal opinion, is that the problem is coming from the server, the response is shaped like this thanks to a flattening operation that only exists in rust. In other word, the serialisation of the real object is not following standard json.

It all depends on how Meilisearch team will consider this issue. If they consider this response ok, then the issue is on our side and we will have to do some akward things here. If they consider that using serd's flatten option is not ideal, then the fix will be very natural here.

ahmednfwela:
I think the flattening is ok here, since following the principle of least surprise, if a user who is used to a list of 
```json
{"id": 1, "name": "Hello world"}
```
suddenly starts seeing a list of 
```json
{"original":{ "id": 1, "name": "Hello world"}, "formatted":{ "id": 1, "name": "Hello world"}}
```
when they request highlighting, it would throw them off.

juchom:
For me, I think that it's really weird to ask a developer to chose between two functions for searching with different return types. 

```csharp
public async Task<ISearchable<T>> SearchAsync<T>(string query,
            SearchQuery searchAttributes = default, CancellationToken cancellationToken = default)
```

and 

```csharp
public async Task<ISearchable<IFormatContainer<T, TFormatted>>> SearchAsync<T, TFormatted>(string query,
            SearchQuery searchAttributes = default, CancellationToken cancellationToken = default)
```

The response should always be the same and the formatted property should always exists as a nullable property.

Otherwise you would have to duplicate some code in your ui depending if you want to see highlighting or not.

The other problem with flattening here, is name collision, what if your model has a `_formatted` property what will be the value of it ?

ahmednfwela:
the second method is just a sugar syntax for the first one actually, because you can just do `SearchAsync<IFormatContainer<T,TFormatted>>(...)`

the developer can keep using `IFormatContainer` whether `_formatted` exists or not, but if they are not going to use highlighting EVER and not willing to keep repeating `.Original` on every property access, I think it's better to use the first overload.

also for name collisions, I think it's normal, as long as it's well-documented it becomes the user's problem to avoid using preserved fields

juchom:
> the developer can keep using `IFormatContainer` whether `_formatted` exists or not, but if they are not going to use highlighting EVER and not willing to keep repeating `.Original` on every property access, I think it's better to use the first overload.

This actual could be solve with this 

```csharp
var item = response.Original;
```

> also for name collisions, I think it's normal, as long as it's well-documented it becomes the user's problem to avoid using preserved fields

Not my opinion, this issue can be avoided without this transformation. Users who are evaluating the option Meilisearch are not going to break what they already have just to make it compatible with it.

I'm sure this is a very low percentage and Meilisearch will have to evaluate what is more interesting for them. This is not only technical, it is now a business decision too.

ahmednfwela:
it isn't all that rare for databases to enforce field naming rules, e.g. in mongodb
image
also users don't break what they have to transform to meili, since meili isn't supposed to be a source of truth, I would like to think of meili as a caching layer (like redis) with improved searching capabilities.

I for instance have a c# server talking to a mongodb instance and syncing data with a meilisearch instance, where my mongodb objects have PascalCase field names and my meilisearch instance has camelCase.
I also transform c# datetimes into unix timestamps for before date/after date filters.

what I am suggesting is keep everything as is server side, and let clients figure it out

juchom:
> what I am suggesting is keep everything as is server side, and let clients figure it out

I'm ok with that if this is the decision of Meilisearch team.

What I was more saying is that before merging this PR, we should wait to know what the core team wants to do with this issue.

This is impacting all strongly typed languages. It's not an easy decision to take for them.

bidoubiwa:
Hey @ahmednfwela!

Just to let you know @alallema is on Holiday and will review this PR once they come back :)

ahmednfwela:
merged with main and solved conflicts @alallema

src/Meilisearch/DefaultFormattable.cs
```diff
@@ -0,0 +1,34 @@
+using System.Text.Json.Serialization;
+
+namespace Meilisearch
+{
+    /// <summary>
+    /// The default implmentation of <see cref="IFormatContainer{TOriginal,TFormatted}"/>
```
alallema:
```suggestion
    /// The default implementation of <see cref="IFormatContainer{TOriginal,TFormatted}"/>
```

tests/Meilisearch.Tests/SearchTests.cs
```diff
@@ -32,6 +33,33 @@ public async Task InitializeAsync()
 
         public Task DisposeAsync() => Task.CompletedTask;
 
+        [Fact]
+        public async Task TestJsonConverter()
```
alallema, line 37: Why is this method asynchronous if there is no `await` inside?

alallema (commented):
Thank you so much @ahmednfwela, for this PR ❤️ and so sorry for the delay ...
I'll make a few comments, but shouldn't we change the samples in the code-sample file according to this new PR?
Also I didn't have the right on this PR to fix the conflicts I let you do it 😃

ahmednfwela:
@alallema Thanks for your review!

I think after the recent changes in meilisearch, it's better if we take an approach similar to this PR https://github.com/meilisearch/meilisearch-dart/pull/350

where we make a comprehensive `MeiliDocumentContainer<T>` class, which wraps around the document and strongly types all the meilisearch-related fields, without making any assumptions on the returned type.

what do you think ?

tests/Meilisearch.Tests/SearchTests.cs
ahmednfwela: I guess it's just a habit of making everything async 😄

alallema:
@ahmednfwela, I more than agree. But I think it's up to @brunoocasali to take the decision

brunoocasali:
> where we make a comprehensive MeiliDocumentContainer<T> class, which wraps around the document and strongly types all the meilisearch-related fields, without making any assumptions on the returned type.

But I think that triggers the same issue, no? Because it may not be possible to do it...

brunoocasali:
bors try

ahmednfwela:
@brunoocasali I am currently working on a refactor for this PR

brunoocasali:
> @brunoocasali I am currently working on a refactor for this PR

Thank you very much @ahmednfwela ;)

curquiza:
@ahmednfwela, thank you again for this PR. DO you think we should continue to work on it?

ahmednfwela:
@curquiza Hi, and sorry for the long delay, but yes, I am currently updating this PR.

danFbach:
@ahmednfwela do you need any assistance with this? 
I like your initial PR, but I had an another idea which may be simpler while providing the same functionality.

create an abstact class
```c#
    /// <summary>
    /// Inheritable class for retrieving formatted results.
    /// </summary>
    /// <typeparam name="T"></typeparam>
    public abstract class MeilisearchDocumentContainer<T>
    {
        /// <summary>
        /// Gets or sets the _formatted field.
        /// </summary>
        [JsonPropertyName("_formatted")]
        public T _Formatted { get; set; }
    }
```
Then if the user sets the "AttributesToHighlight" and are expecting `_formatted` to be returned, they just define their class as inheriting from `MeilisearchDocumentContainer`.

This method passes all existing tests without modification (other than the `Movie` class inheriting from `MeilisearchDocumentContainer` and replacing the `FormattedMovie` with Movie that inherits from `MeilisearchDocumentContainer`) and gives users the ability to easily use the `_formatted` feature and would be a non-breaking change.

Let me know your thoughts!

ahmednfwela:
hi @danFbach , my main problem with inheritance, is that we are restricting the user to use our base class, so if they have their own base class for dtos, they will be unusable 

you are free to pick up the work here though, as I am currently extremely busy, there is a similar implementation in meilisearch-dart repo if you want it for reference

PR #409, opened 2023-03-29
Add FacetStats field in SearchResponse

alallema:
As per [the specification](https://github.com/meilisearch/specifications/pull/224)
SDK requirements: https://github.com/meilisearch/integration-guides/issues/251

A new `facetStats` field is returned when `facets` is used in the search parameters and contains numerical values. The min and max numeric values of these facets will be returned.

### DIFF
--- a/src/Meilisearch/ISearchable.cs
+++ b/src/Meilisearch/ISearchable.cs
@@ -38,5 +38,11 @@ namespace Meilisearch
         /// </summary>
         [JsonPropertyName("_matchesPosition")]
         IReadOnlyDictionary<string, IReadOnlyCollection<MatchPosition>> MatchesPostion { get; }
+
+        /// <summary>
+        /// Returns the numeric min and max values per facet of the hits returned by the search query.
+        /// </summary>
+        [JsonPropertyName("facetStats")]
+        IReadOnlyDictionary<string, IReadOnlyDictionary<string, float>> FacetStats { get; }
     }
 }
--- a/src/Meilisearch/ISearchable.cs
+++ b/src/Meilisearch/ISearchable.cs
@@ -43,6 +43,6 @@ namespace Meilisearch
         /// Returns the numeric min and max values per facet of the hits returned by the search query.
         /// </summary>
         [JsonPropertyName("facetStats")]
-        IReadOnlyDictionary<string, IReadOnlyDictionary<string, float>> FacetStats { get; }
+        IReadOnlyDictionary<string, FacetStat> FacetStats { get; }
     }
 }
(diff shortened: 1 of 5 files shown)


src/Meilisearch/PaginatedSearchResult.cs
```diff
@@ -18,7 +18,8 @@ public class PaginatedSearchResult<T> : ISearchable<T>
             IReadOnlyDictionary<string, IReadOnlyDictionary<string, int>> facetDistribution,
             int processingTimeMs,
             string query,
-            IReadOnlyDictionary<string, IReadOnlyCollection<MatchPosition>> matchesPostion
+            IReadOnlyDictionary<string, IReadOnlyCollection<MatchPosition>> matchesPostion,
+            IReadOnlyDictionary<string, IReadOnlyDictionary<string, float>> facetStats
```
bidoubiwa: Do you think it's a good idea to specify the fields `min: float` and `max: float` instead of `<string, float>` since we know what it contains?

alallema: Actually, we know that it will contain an array of facets with min and max in it, but we don't know which facet. Or maybe I didn't understand well

bidoubiwa: I'm not very good with dotnet but I suppose something like this:

```
public class FacetStat
{
    public int min { get; set; }
    public int max { get; set; }
}
IReadOnlyDictionary<string, FacetStat> facetStats
```

Where string is the `facet` we do not know about, and `facetStat` its value
```
{
   // <string, FacetStat>
   "genres": { min: 1, max: 2 }
}
```

alallema: Of course! I didn't understand that at all

src/Meilisearch/ISearchable.cs
```diff
@@ -43,6 +43,6 @@ public interface ISearchable<T>
         /// Returns the numeric min and max values per facet of the hits returned by the search query.
         /// </summary>
         [JsonPropertyName("facetStats")]
-        IReadOnlyDictionary<string, IReadOnlyDictionary<string, float>> FacetStats { get; }
+        IReadOnlyDictionary<string, FacetStats> FacetStats { get; }
```
bidoubiwa: There is a conflict as you are using both times `facetStats`. Is this an issue? Otherwise you can name the class `facetStat` and the other `facetStats`

alallema: No is not a problem but you right actually is just one facet stat so it should be named `FacetStat`

bidoubiwa (approved):
🔥

PR #438, opened 2023-06-27
Introduce Get/delete documents by filter (Meilisearch v1.2.0)

### LINKED ISSUE #431: Delete documents by filter
⚠️ This issue is generated, it means the examples and the namings do not necessarily correspond to the language of this repository.
Also, if you are a maintainer, feel free to add any clarification and instruction about this issue.

Sorry if this is already partially/completely implemented, feel free to let me know about the state of this issue in the repo.

Related to https://github.com/meilisearch/integration-guides/issues/261

----

This issue is divided in two sections, first you need to make the implementation, second you must update the code-samples (no one likes a outdated docs, right?).

### New implementation 

Related to:
- issue: https://github.com/meilisearch/meilisearch/issues/3477
- spec: https://github.com/meilisearch/specifications/pull/236

It allows the user to use a filter expression to remove documents.

Since there is already a method called `deleteDocuments(uids)` in the SDKs, and this method is calling the route `POST /indexes/:index_uid/documents/delete-batch`, the requirements to implement the new way is to create a conditional internally to check the presence of the new parameter `filters` and then call the new method `POST /indexes/:index_uid/documents/delete`.
When the developer uses the old `ids` argument, it should keep calling the old implementation.

This will avoid any **silent failures** the users may find.

For example, in the Ruby SDK:

```rb
# Public: Delete documents from an index
#
# documents_ids - An array with document ids (deprecated, optional)
# filter - A hash containing a filter that should match documents. Available ONLY with Meilisearch v1.2 and newer (optional)
# 
# Returns a Task object.
def delete_documents(documents_ids = nil, filter: nil)
  if documents_ids
    http_post "/indexes/#{@uid}/documents/delete-batch", documents_ids
  else
    http_post "/indexes/#{@uid}/documents/delete", { filter: filter }
  end
end
```

**Extra:** Add inline documentation for the method, explaining the availability of the filter syntax only for Meilisearch v1.2 and newer.

**Extra:** Mark the `document_ids` parameter as `deprecated`. eg. `@Deprecated('migration')` on Java/Dart, Obsolete on C#, etc. Add a library if necessary.

**Extra:** Add a `try`/`catch` to detect the error and give the user a hint about the possibility of a version mismatch between the SDK version and the instance version. Check this PR for an example in JavaScript: https://github.com/meilisearch/meilisearch-js/pull/1492

**Extra:** General recommendations:
- For languages where method overloading is available, use the overloading.
- For languages where a builder is available, use the builder pattern.

### Code samples:

Inside of this file: `.code-samples.meilisearch.yml`:

1. Rename this key: `delete_documents_1:` to `delete_documents_by_batch_1:`
2. Create a new entry with this key `delete_documents_by_filter_1:` containing a call to the `deleteDocuments` method 
using the new behavior from the index `movies` using the `filter` param with this string `"genres = action OR genres = adventure"`.

Use this as a reference _if the previous description was not helpful_:
```
  POST http://localhost:7700/indexes/movies/documents/delete
  with data: { "filter": "genres = action OR genres = adventure" }
```

### Todo:

- [ ] Implement the `deleteDocuments` new behavior keeping the old behavior when possible.
- [ ] Update the code-samples according to this issue.

### LINKED ISSUE #432: Get documents by filter
⚠️ This issue is generated, it means the examples and the namings do not necessarily correspond to the language of this repository.
Also, if you are a maintainer, feel free to add any clarification and instruction about this issue.

Sorry if this is already partially/completely implemented, feel free to let me know about the state of this issue in the repo.

Related to https://github.com/meilisearch/integration-guides/issues/261

----

This issue is divided in two sections, first you need to make the implementation, second you must update the code-samples (no one likes a outdated docs, right?).

### New implementation 

Related to:
- issue: https://github.com/meilisearch/meilisearch/issues/3477
- spec: https://github.com/meilisearch/specifications/pull/234

Gives the user the possibility to use a filter expression to retrieve documents, like in `search`.

Implement in the `getDocuments` method an internal conditional allowing the user to query the documents with the same method but using a new filter method.

When the user calls the get documents method with a `filter` argument, request `POST /indexes/{index_uid}/documents/fetch` using a JSON body containing an `int offset`, `int limit`, `String[] fields` and now `String filter` or `String[] filter`.
⚠️ If the method invocation does not contain a `filter` it should still call the previous implementation.

### Code samples:

Inside of this file: `.code-samples.meilisearch.yml`:

1. Replace this key: `get_documents_1:` content to use the `filter` `genres=action`, keep the `index name` and the `limit` just add the `filter`.
Use this as a reference _if the previous description was not helpful_::
```
get_documents_1: |- 
  GET 'http://localhost:7700/indexes/movies/documents?limit=2&filter=genres=action
  mind the filter in the query param (use the POST route in the integrations)
```

2. Create a new entry with this key `get_documents_post_1:` containing a call to the `getDocuments` method 
using the new behavior from the index `movies` using the `filter` param with this string `"genres = action OR genres = adventure"`.

Use this as a reference _if the previous description was not helpful_:
```
get_documents_post_1: |-
  POST http://localhost:7700/indexes/books/documents/fetch
  with data: { 
    "filter": "(rating > 3 AND (genres = Adventure OR genres = Fiction)) AND language = English",
    "fields": ["title", "genres", "rating", "language"],
    "limit": 3
  }
```

### Todo:

- [ ] Implement the `getDocuments` new behavior keeping the old behavior when possible.
- [ ] Update the code-samples according to this issue.

ahmednfwela:
# Pull Request

## Related issue
Fixes #432
Fixes #431

## What does this PR do?
- Adds `object Filter` Property to `DocumentsQuery`
- Adds `Task<TaskInfo> DeleteDocumentsAsync(DeleteDocumentsQuery query, CancellationToken cancellationToken = default)` to Index
- Update .env files to use meilisearch 1.2.0
- The rest is just formatting changes

## PR checklist
- [x] Add the changes in code samples

cc @alallema @brunoocasali


src/Meilisearch/Index.Documents.cs
```diff
@@ -430,6 +453,25 @@ public async Task<T> GetDocumentAsync<T>(int documentId, List<string> fields = d
...
         }
 
+        /// <summary>
+        /// Delete documents from an index based on a filter.
+        /// </summary>
+        /// <remarks>Available ONLY with Meilisearch v1.2 and newer.</remarks>
+        /// <param name="query">A hash containing a filter that should match documents.</param>
+        /// <param name="cancellationToken">The cancellation token for this call.</param>
+        /// <returns>Return the task info.</returns>
+        public async Task<TaskInfo> DeleteDocumentsAsync(DeleteDocumentsQuery query,
```
alallema, line 472:
```suggestion
        public async Task<TaskInfo> DeleteDocumentsAsync(DocumentsQuery query,
```
You don't have to create a new class, you can use the existing `DocumentsQuery` WDYT?

tests/Meilisearch.Tests/DocumentTests.cs
```diff
@@ -134,10 +135,8 @@ public async Task BasicDocumentsAdditionInBatches()
             // Add the documents
             Movie[] movies =
             {
-                new Movie { Id = "1", Name = "Batman" },
-                new Movie { Id = "2", Name = "Reservoir Dogs" },
-                new Movie { Id = "3", Name = "Taxi Driver" },
-                new Movie { Id = "4", Name = "Interstellar" },
+                new Movie { Id = "1", Name = "Batman" }, new Movie { Id = "2", Name = "Reservoir Dogs" },
+                new Movie { Id = "3", Name = "Taxi Driver" }, new Movie { Id = "4", Name = "Interstellar" },
```
alallema, line 139: Is it because of the linter because I find it less legible on several lines? Especially for tests

tests/Meilisearch.Tests/DocumentTests.cs
```diff
@@ -419,10 +418,8 @@ public async Task BasicDocumentsUpdateInBatches()
...
```
alallema, line 422: Same here

.code-samples.meilisearch.yaml
```diff
@@ -80,7 +80,13 @@ delete_an_index_1: |-
 get_one_document_1: |-
   await client.Index("movies").GetDocumentAsync<Movie>(25684, new List<string> { "id", "title", "poster", "release_date" });
 get_documents_1: |-
-  await client.Index("movies").GetDocumentsAsync<Movie>(new DocumentsQuery() { Limit = 2 });
+  await client.Index("movies").GetDocumentsAsync<Movie>(new DocumentsQuery() { Limit = 2, Filter = "genres = action" });
+get_documents_post_1: |-
+  await client.Index("movies").GetDocumentsAsync<Movie>(new DocumentsQuery() { 
+      Limit = 3, 
+      Fields = new List<string> { "title", "genres", "rating", "language"}, 
+      Filter = "genres = action OR genres = adventure"
```
alallema:
```suggestion
      Filter = "(rating > 3 AND (genres=Adventure OR genres=Fiction)) AND language=English" 
```
There is a mistake in the issue sorry about that

alallema (commented):
Thank you so much for this PR and for contributing to Meilisearch @ahmednfwela ❤️ 
I've added a few comments.
Also, if you have the time and, above all, the inclination, we've added a method to all SDKs when adding methods name `VersionErrorHintMessage` that warns the user if an error occurs. I'll let you check the discussion [about it](https://github.com/meilisearch/integration-guides/issues/261#issuecomment-1551072528). And, of course, no obligation I will create an issue about it later in the case.
Don't hesitate if you have any questions 😊

src/Meilisearch/Index.Documents.cs
ahmednfwela: We can't here since `DocumentsQuery` contains more properties than `DeleteDocumentsQuery`

tests/Meilisearch.Tests/DocumentTests.cs
ahmednfwela: Nope, this was because I am using a custom formatter that sets line limits, and this was just it reformatting on saving the document.
apologies about that !

alallema (approved):
LGTM! 🔥 🔥 🔥 
Thanks again @ahmednfwela for this PR

alallema:
Hi @ahmednfwela I'm not able to update your branch before merge it if you want to do it ❤️

ahmednfwela:
@alallema should be ok now 👍

ahmednfwela:
I have no idea why the CI is failing, probably a flakey test ?

alallema:
> I have no idea why the CI is failing, probably a flakey test ?

I have no idea either ... Sorry for the delay

alallema:
bors try

alallema:
bors merge

PR #578, opened 2024-11-11
fix for #577

### LINKED ISSUE #577: EstimatedTotalHits Removed
**Description**
The Search was refactored at some point to use https://github.com/meilisearch/meilisearch-dotnet/blob/main/src/Meilisearch/ISearchable.cs instead of https://github.com/meilisearch/meilisearch-dotnet/blob/main/src/Meilisearch/SearchResult.cs and some properties were left out, like **EstimatedTotalHits**

danFbach:
# Pull Request

## Related issue
Fixes #577 

## What does this PR do?
- brings the `Offset`, `Limit` and `EstimatedTotalHits` fields to `ISearchable<T>`
- migrated fields in `SearchResult<T>` now use `/// <inheritdoc/>` as docs were also moved to `ISearchable<T>`

## PR checklist

Not sure if `Offset` and `Limit` also need to ported up, but i could envision a scenario where it would be useful, so I brought them along too.

### DIFF
--- a/src/Meilisearch/ISearchable.cs
+++ b/src/Meilisearch/ISearchable.cs
@@ -23,6 +23,24 @@ namespace Meilisearch
         IReadOnlyCollection<T> Hits { get; }
 
         /// <summary>
+        /// Number of documents skipped.
+        /// </summary>
+        [JsonPropertyName("offset")]
+        int Offset { get; }
+
+        /// <summary>
+        /// Number of documents to take.
+        /// </summary>
+        [JsonPropertyName("limit")]
+        int Limit { get; }
+
+        /// <summary>
+        /// Gets the estimated total number of hits returned by the search.
+        /// </summary>
+        [JsonPropertyName("estimatedTotalHits")]
+        int EstimatedTotalHits { get; }
+
+        /// <summary>
         /// Returns the number of documents matching the current search query for each given facet.
         /// </summary>
         [JsonPropertyName("facetDistribution")]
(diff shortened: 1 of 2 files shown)


migueltarga (approved):
Awesome!

danFbach:
I think I would retract this PR. There should probably be separate interfaces for offset and paginated results. This would better conform with API result object, which is polymorphic based on the request. Is this more what you all would be looking for?

migueltarga:
The way you have it looks completely fine to me:
https://www.meilisearch.com/docs/reference/api/search#response

danFbach:
@migueltarga The problem is that they already have both `SearchResult<T>` and `PaginatedSearchResult<T>` as implementations of `ISearchable<T>` This _should_ enable you to do something like this...

For paginated results.
```cs
var results = await index.SearchAsync<MeilisearchResultProductModel>(query, new SearchQuery()
{
    ShowRankingScore = true,
    Page = page,
    HitsPerPage = pageSize,
    Filter = buildFilter(categoryIds, manufacturerId, currentStoreId),
});

if (results is PaginatedSearchResult<MeilisearchResultProductModel> paginatedResult)
{
    var totalPages = paginatedResult.TotalPages
}
```

Or for offset style request:
```cs

if (results is SearchResult<MeilisearchResultProductModel> searchResult)
{
    var totalHits = searchResult.EstimatedTotalHits;
}
```

Will be testing this shortly and will report if it works as expected.

danFbach:
@migueltarga I can confirm that the following syntax is valid.
```cs
var results = await index.SearchAsync<T>(query, new SearchQuery()
{
    Limit = 5,
    Offset = 0,
    ShowRankingScore = true
});

if (results is SearchResult<T> resultModel)
{
    var estimatedTotalHits = resultModel.EstimatedTotalHits;
}
```
Docs should probably be updated to make this more obvious.

migueltarga:
@danFbach Thank you for the explanation and the example!

PR #752, opened 2026-07-30
Add personalized search support to the search query

### LINKED ISSUE #737: [Meilisearch v1.47.0] Add personalized search (experimental)
## Context

[Meilisearch 1.47.0](https://github.com/meilisearch/meilisearch/releases/tag/v1.47.0) adds full support for personalized search (experimental) to the search endpoint.

> NB: before 1.47.0, federated search requests did not support personalized search.

## Objective

Update the SDK to support [personalized search](https://www.meilisearch.com/docs/capabilities/personalization/getting_started/personalized_search) on the search endpoint.

Update the corresponding SDK methods
- [Search (POST)](https://www.meilisearch.com/docs/reference/api/search/search-with-post)
- [Search (GET)](https://www.meilisearch.com/docs/reference/api/search/search-with-get)

Both related API endpoints now accept a new `personalize` object parameter:

```json
{
  // other request parameters...
  "personalize": {
    "userContext": "The user prefers ..."
  }
}
```

## Tasks

- [ ] Add/update methods to handle interacting with the search API endpoints
- [ ] Add/update test cases for the new/updated methods
- [x] No code samples to update

arcusbuilds:
# Pull Request

## Related issue

Fixes #737

## What does this PR do?

Adds the experimental `personalize` search parameter from Meilisearch 1.47.0. It is a new `Personalize` class with a single `UserContext` string, exposed as `SearchQuery.Personalize`:

```csharp
await index.SearchAsync<Movie>("man", new SearchQuery
{
    Personalize = new Personalize { UserContext = "The user only watches science fiction movies" }
});
```

This covers Search (POST). Search (GET) has no surface in this SDK: the only search call is the POST in `Index.Documents.cs`, and there is no GET-based search method to update. Worth flagging for whoever adds one, the GET endpoint takes a flat `personalizeUserContext` string rather than a `personalize` object, so this class will not map onto it directly.

Non-federated `/multi-search` picks the parameter up without extra work, since `MultiSearchQuery.Queries` is a `List<SearchQuery>`.

### Not included

Federated `/multi-search` needs `federation.personalize` instead. The server rejects per-query `personalize` when a `federation` object is present, and `FederatedMultiSearchQuery` always emits one, so federated personalization is currently unreachable from the SDK. I kept this PR scoped to the issue and left it out. Happy to open a separate PR for it if you want.

### Tests

Reranking needs a personalization API key on the instance, which neither `docker-compose.yml` nor CI sets, so the feature itself cannot run under test. Two things are covered instead:

- `SearchQuerySerializationTests` asserts the request JSON directly, with no server involved.
- `CustomSearchWithPersonalize` sends a real request and passes whether or not a key is configured, so it will not start failing if CI ever gets one.

All of the tests added here pass locally against Meilisearch 1.51.0.

## PR checklist

- [x] Did you use any AI tool while implementing this PR (code, tests, docs, etc.)? If yes, disclose it in the PR description and describe what it was used for. AI usage is allowed when it is disclosed.

### AI tool disclosure

I used Claude Code to write the implementation and the tests, and to review the result. I verified the API shape by hand against a local Meilisearch 1.51.0 instance, including which error code comes back for each request shape, and I read through the whole diff before committing.

## Summary by CodeRabbit

- **New Features**
  - Added support for personalizing search queries using a user context.
  - Personalization can now be included in standard and multi-search requests.
  - Requests use the `personalize.userContext` format for compatibility with the search API.

- **Tests**
  - Added coverage for personalization serialization, omission when unused, multi-search requests, and search behavior when personalization is enabled.


arcusbuilds:
@curquiza @Strift this has been open a month with CI green. It adds the `personalize` parameter from #737, covering search and non-federated multi-search. Could one of you take a look?

arcusbuilds:
@Strift Following up on this once more. I'm happy to rebase or change the approach if the API has moved since #737. If personalized search isn't something you want in the SDK right now, just let me know and I'll close the PR.

PR #758, opened 2026-08-11
Add `indexSize` and `usedIndexSize` to index stats

### LINKED ISSUE #757: [Meilisearch v1.53.0] Add `indexSize` and `usedIndexSize` to index stats
## Context

[Meilisearch v1.53.0](https://github.com/meilisearch/meilisearch/releases/tag/v1.53.0) adds `indexSize` and `usedIndexSize` to per-index stats.

## Objective

Update the SDK so index stats responses include the new **`indexSize`** and **`usedIndexSize`** fields returned by [Get stats of index](https://www.meilisearch.com/docs/reference/api/indexes/get-stats-of-index) and per index in [Get stats of all indexes](https://www.meilisearch.com/docs/reference/api/stats/get-stats-of-all-indexes).

- `indexSize`: size of the index DB (bytes by default; honour `sizeFormat` when supported)
- `usedIndexSize`: size of the used pages of the index DB (bytes by default; honour `sizeFormat` when supported)

## Tasks

- [ ] Add `indexSize` and `usedIndexSize` to the index stats response shape
- [ ] Update test cases

ahmetbahadiryilmaz:
## Summary

Closes #757.

Adds the `indexSize` and `usedIndexSize` fields introduced in [Meilisearch v1.53.0](https://github.com/meilisearch/meilisearch/releases/tag/v1.53.0) (meilisearch/meilisearch#6563) to `IndexStats`. They are therefore available from both `index.GetStatsAsync()` and per index in `client.GetStatsAsync()`.

## Why the properties are `long?` rather than `long`

`IndexStats` is bound by `System.Text.Json` through its constructor. When a JSON property is absent, the corresponding parameter silently receives its default — so a non-nullable `long IndexSize` would report **`0`** against any Meilisearch older than v1.53.0, which is indistinguishable from an index that genuinely occupies very little space. This client supports servers well below that version; the test suite here currently runs against `v1.16`.

Making them `long?` keeps that distinction honest: `null` means "this instance does not report the metric", `0` means zero. Callers that do not care can write `stats.IndexSize ?? 0`.

Both constructor parameters are optional and appended at the end, so existing calls are unaffected at the source level.

I am happy to switch them to non-nullable `long` for consistency with `RawDocumentDbSize` and `AvgDocumentSize` if you would rather keep the shape uniform — just say the word.

## Tests

The existing stats coverage is integration-only (`IndexTests.GetStats` just asserts the response is not null), and CI runs a Meilisearch version that does not yet return these fields, so I added `tests/Meilisearch.Tests/IndexStatsTests.cs` — three deserialization tests that need no running server:

- new fields are read when the instance reports them,
- they are `null` when the instance omits them, while the surrounding fields still bind correctly,
- they are also read from the per-index payload of the all-indexes `Stats` response.

Verified locally:

| Check | Result |
| --- | --- |
| `dotnet build --configuration Release` | succeeded, no new warnings |
| `dotnet test --filter FullyQualifiedName~IndexStatsTests` | 3 passed, 0 failed |
| `dotnet format --verify-no-changes` | clean |

I did not touch `MEILISEARCH_VERSION` in `.env`, since bumping the tested server version seems like a separate concern. Once it reaches v1.53.0, `IndexTests.GetStats` could additionally assert the sizes are non-null.

---

_AI disclosure, per the "Coding with AI" section of `CONTRIBUTING.md`: I used Claude Code while working on this — to explore the codebase, draft the implementation and the tests, and draft this description. I reviewed all of it, made the nullability decision described above, and verified it locally with the three commands in the table._

## Summary by CodeRabbit

* **New Features**
  * Index statistics now include total and used index sizes when reported by the server.
  * Index-size values are available in both individual and aggregate statistics.

* **Bug Fixes**
  * Missing index-size data is handled gracefully, remaining unavailable when unsupported.
  * Reported zero sizes are preserved accurately instead of being treated as missing.


ahmetbahadiryilmaz:
﻿Thanks — both points addressed in the follow-up commit.

**Binary compatibility:** you were right, adding optional parameters to the existing public constructor changes its signature and would break already-compiled consumers with a `MissingMethodException`. The seven-parameter constructor is now preserved and delegates to the nine-parameter one, which carries `[JsonConstructor]` so `System.Text.Json` picks it unambiguously. I dropped the default values from the nine-parameter overload, since keeping them alongside the seven-parameter one would have made a seven-argument call ambiguous at compile time.

**Explicit zero coverage:** added `ReportedZeroSizesAreDistinctFromOmittedOnes`, which asserts that `"indexSize": 0` deserializes to `0` rather than `null` — that is exactly the distinction the nullable type exists to preserve, so it deserved a test. Also added `ConstructorWithoutIndexSizesLeavesThemNull` to pin the preserved constructor.

Local run is now 5 passed / 0 failed, with `dotnet format --verify-no-changes` and the Release build still clean.

